PR #30, opened 2022-04-13
Be conformant with flags to xa_end

### LINKED ISSUE #28: Be conformant with flags to xa_end
The XA spec explicitly states that:

> Following are the valid settings of flags (note that one and only one of TMSUSPEND,
TMSUCCESS or TMFAIL must be set):

I don't think that we want to set `TMFAIL` since this wold make it impossible for the user to keep using resources until the mandatory upcoming rollback. The specs explicitly states that it's ok to use resources after a transaction is put in _abort only_, but the work will of corse be rolled back.

lazan:
Make sure we always uses exactly on of [ flag::xa::Flag::success,
flag::xa::Flag::suspend, flag::xa::Flag::fail] on xa_end.

Right now we never set _fail_ since this would make it impossible to
interact with the resource after. All units of work will be rolled back,
but the user might want to interact with resources for reads after a
service has return with _fail_. Well, it's possible a quality of
implementation issue for the resource, but at least db2 makes it impossible...

fixes #28

### DIFF
--- a/middleware/common/include/common/flag.h
+++ b/middleware/common/include/common/flag.h
@@ -49,9 +49,9 @@ namespace casual
          constexpr Flags() = default;
 
          template< typename... Enums>
-         constexpr Flags( enum_type e, Enums... enums) : Flags( bitmask( e, enums...)) {}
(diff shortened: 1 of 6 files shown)


ahhud (approved):
Looking good.

PR #33, opened 2022-04-17
cli - fix confusing arguments for casual service --legend

albinsjolund:
Before:
--legend triggerd on list-service

After:
--legend triggers on list-services. list-service has been removed from --help but still triggers --legend for backward compatibility


lazan (approved):
I was hoping for a unittest also, but I don't think we need it for this change, it't pretty trivial :)
We do need to start accumulating unittest for cli in general though.

lazan:
Use _Rebase and merge_ to merge the PR

lazan:
Merge "by hand" - need to remove a few commits from 1.5.11

PR #45, opened 2022-05-09
Feature/1.5/patch.12 gateway outbond transaction mappings are growing over time

lazan:
@OlofGullnas These two commits are probably of interest to BV, please review them.

It is probably of value if both @albinsjolund and @AlbsAgren take a peek at the commits as well.

@ahhud, if you got time.

### DIFF
--- a/middleware/common/source/unittest/rm.cpp
+++ b/middleware/common/source/unittest/rm.cpp
@@ -163,7 +163,7 @@ namespace casual
                return state.error( code::xa::protocol);
             }
 
-            log::line( log, "xa_open_entry - openinfo: ", openinfo, " rmid: ", rmid, " flags: ", flags);
+            log::line( log, "xa_open_entry - openinfo: ", openinfo, ", rmid: ", rmid, " flags: ", flags);
             return cast::underlying( state.result.open);
          }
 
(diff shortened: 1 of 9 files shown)


ahhud (approved):

OlofGullnas (approved):
Seems reasonable, but got me thinking.

I still do not have a gut feeling for how transaction handling is implemented and works in Casual. I have not really looked into how it works vhen multiple domains (gateways) are involved.

What is  the rule for when a one phase commit optimization may be used? i.e. a simple commit without a prepare phase...
 
I think it is: one (1) and only one resource is used in exactly one transaction branch in the "complete transaction tree."

In theory there can be cases where the transaction (branch) in the "root domain" has zero resources associated with it and a remote domain has a single resource associated with the global transaction. A  one phase optimization could possibly be used in this case, but would require "delegation" of the commit to the remote domain transaction manager. (This might be impossible to do for other reasons...) 

Another unlikely scenario might be exactly one RM in the "root transaction branch" and no RMs used in remote domains involved in the transaction. Also very unlikely, but not impossible Possibility may increase when dynamic registration is used. 
  
Theese are unusual scenarios, so it is doubtful if it is worth considering them! It probably also requires complete knowledge about ALL transaction branches in the root domain transaction manager. It can never be wrong to use the two-phase commit, so 2 phase is always safe.  

Keeping the decision simple is good :-)

PR #48, opened 2022-05-15
Feature/1.5/patch.12 service use ipc multiplex send everywhere we send

AlbsAgren:


middleware/service/source/forward/handle.cpp
```diff
@@ -92,7 +92,7 @@ namespace casual
 
                         log::line( log, "send request - to: ", message.process, " - request: ", request);
 
-                        if( ! communication::ipc::flush::optional::send( message.process.ipc, request))
+                        if( ! state.multiplex.send( message.process.ipc, request))
```
lazan: What if the caller "is gone" later? Hmm, we would not send an error reply...

middleware/service/source/forward/main.cpp
```diff
@@ -41,17 +42,35 @@ namespace casual
...
+               {
+                  return [ handler = handle::create( state)]( common::strong::file::descriptor::id descriptor, common::communication::select::tag::read) mutable
+                  {
+                     auto& inbound = common::communication::ipc::inbound::device();
+                     if( inbound.descriptor() != descriptor)
+                        return false;
+
+                     handler( common::communication::device::non::blocking::next( inbound));
+                     return true;
+                  };
```
lazan, line 57: Nice! We may want to consume _next message_ until device is depleted, or we've reached some _platform::batch::..limit_ (low, under 20 maybe). Not that important in the forward I'm guessing. Not that much load...

middleware/service/source/manager/handle.cpp
```diff
@@ -58,44 +58,16 @@ namespace casual
...
-                        return false;
-                     }
-
-                  }
-               } // eventually
-
                namespace error
                {
-                  auto reply( const state::instance::Caller& caller, common::code::xatmi code)
+                  auto reply( State& state, const state::instance::Caller& caller, common::code::xatmi code)
```
lazan, line 70: A given extension to the "api", nice.

middleware/service/source/manager/handle.cpp
```diff
@@ -427,7 +399,7 @@ namespace casual
                               }                             
 
                               // send reply, if caller gone, we discard the reservation.
-                              if( ! local::optional::send( message.process.ipc, reply))
+                              if( ! local::optional::send( state, message.process.ipc, reply))
```
lazan: What if the caller "is gone" later? Hmm, we would not discard the reservation.

middleware/service/include/service/forward/state.h
```diff
@@ -37,10 +39,18 @@ namespace casual
       
       struct State
       {
+         State()
```
lazan:
```cpp
inline State() ...
```

middleware/service/source/manager/handle.cpp
```diff
@@ -58,44 +58,16 @@ namespace casual
...
-                        auto error = exception::capture();
-                        if( error.code() != code::casual::communication_unavailable)
-                           throw;
-                           
-                        log::line( log, error, " destination unavailable - ", destination);
-                        return false;
-                     }
-
-                  }
-               } // eventually
```
lazan, line 94: The best change! Removal of code...

lazan (changes requested):
We need to do something about if "targets" are unavailable later during multiplex send. Two options:
* Use "flush::send" in these cases, not attractive...
* Add a `common::unique_function` callable to the multiplex send state, and if "target" is unavailable AND  _callable_ is not 'null', call it (to send error message and such..

lazan:
Also: The commit header messages should only refer to `service -  ...`
* `service - use multiplex send in manager`.   
* `service - use multiplex send in forward`.
Or something similar. The first _word_ is the _module_, and we don't use finer granularity, if we don't need to. 
You could squash these two also.

lazan:
Somethings must have gone wrong with the "rebase from feature/1.5/patch.12". You might need to move this to another branch, take out a new branch from origin/feature/1.5/patch.12  and cherry-pick your commits.

AlbsAgren:
What doesn't look right? I've squashed my commits together and rebased off of feature/1.5/patch.12 (though I see now that there's been another pr merged since I rebased). Other than that, the diff looks correct to me as far as I can tell.

lazan:
My bad! You're totally right, sorry.

middleware/service/source/forward/handle.cpp
```diff
@@ -92,12 +92,11 @@ namespace casual
 
                         log::line( log, "send request - to: ", message.process, " - request: ", request);
 
-                        if( ! communication::ipc::flush::optional::send( message.process.ipc, request))
+                        state.multiplex.send( message.process.ipc, request, [ &state, &request, service_name = message.service.name]( auto& destination, auto& complete)
```
lazan: The request that you capture by ref, won't be alive when the lambda is invoked. Pretty much all things beside `State` must be capture by value.

middleware/service/source/manager/handle.cpp
```diff
@@ -427,10 +399,10 @@ namespace casual
                               }                             
 
                               // send reply, if caller gone, we discard the reservation.
-                              if( ! local::optional::send( message.process.ipc, reply))
+                              state.multiplex.send( message.process.ipc, reply, [ &state, &handle]( auto& destination, auto& complete){
```
lazan: `handle` captured by ref. capture pid by value.

lazan (changes requested):

middleware/service/source/forward/handle.cpp
AlbsAgren: Of course, too much javascript on the brain.

lazan: :)

lazan (approved):
Seams ok. Good work. If you merge it your self (I think you can) use _rebase merge_ and be aware that GitHub change the sha-sum in your commits, so your local branch won't be "equal".

PR #54, opened 2022-05-30
transaction - reply with read_only if an ongoing transaction gets an external commit/prepare::Request

albinsjolund:
Before:
Transaction manager sometimes tried to commit stuff that had already been handled
when an external resource sent a request.

### DIFF
--- a/middleware/transaction/unittest/source/test_manager.cpp
+++ b/middleware/transaction/unittest/source/test_manager.cpp
@@ -1634,6 +1634,262 @@ domain:
          // unittest only...
          common::transaction::context().clear();
       }
+
+      TEST( transaction_manager, two_resources__send_prepare_request__one_external_resource__send_one_phase_commit__expect_ok)
+      {
+         common::unittest::Trace trace;
+
+         auto domain = local::domain( local::configuration::system, local::configuration::base);
+
+         local::involved::Process rm1;
+         local::involved::Process rm2;
+
+         auto trid = common::transaction::id::create( process::handle());
+         auto branch = common::transaction::id::branch( trid);
+
+         // involve resource 1 and 2, resource 2 is an outbound
+         {
+            common::message::transaction::resource::external::Involved message;
+            
+            message.process = rm1.process;
+            message.trid = trid;
+            local::send::tm( message);
+
+            message.process = rm2.process;
+            message.trid = branch;
+            local::send::tm( message);
+         }
+
+         // act as "user" and send commit::Request to transaction manager
+         {
+            common::message::transaction::commit::Request message;
+            message.trid = trid;
+            message.process = process::handle();
+
+            local::send::tm( message);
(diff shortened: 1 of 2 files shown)


middleware/transaction/unittest/source/test_manager.cpp
```diff
@@ -1634,6 +1634,262 @@ namespace casual
...
+            common::message::transaction::resource::external::Involved message;
+            
+            message.process = rm1.process;
+            message.trid = trid;
+            local::send::tm( message);
+
+            message.process = rm2.process;
+            message.trid = branch;
+            local::send::tm( message);
+         }
```
lazan, line 1661: This is not just "gateway involved"

middleware/transaction/unittest/source/test_manager.cpp
```diff
@@ -1634,6 +1634,262 @@ namespace casual
...
+
+         // send commit::Request
+         {
+            common::message::transaction::commit::Request message;
+            message.trid = trid;
+            message.process = process::handle();
+
+            local::send::tm( message);
+         }
```
lazan, line 1670: What does "commit::Request" mean? I think you need to describe that you act as "user"/tx_begin()

middleware/transaction/unittest/source/test_manager.cpp
```diff
@@ -1634,6 +1634,262 @@ namespace casual
...
+            communication::device::blocking::receive( rm1.inbound, message);
+
+            auto reply = common::message::reverse::type( message);
+
+            reply.resource = message.resource;
+            reply.state = common::code::xa::ok;
+            reply.trid = message.trid;
+
+            local::send::tm( reply);
+         }
```
lazan, line 1685: ... same here.

middleware/transaction/unittest/source/test_manager.cpp
```diff
@@ -1634,6 +1634,262 @@ namespace casual
...
+            }
+
+            auto reply = common::message::reverse::type( message);
+
+            reply.resource = message.resource;
+            reply.state = common::code::xa::ok;
+            reply.trid = message.trid;
+
+            local::send::tm( reply);
+         }
```
lazan, line 1719: I think you need to explain what this is, like "act like another domains TM that got the resource::prepare::Request ....." and so on.
The comment now doesn't really say anything.

lazan (approved):
I approve this, but I really think you need to document/comment the unittests better from a logical perspective. Like: "we act as resource 1, and receive ...bla bla " "we act as outbound (resource) and then send ... bla bla"

middleware/transaction/unittest/source/test_manager.cpp
lazan: Perfekt!

lazan (approved):
Much better!

PR #57, opened 2022-06-09
configuration - fixed issue where configuration normalize returned wrong configuration model

TobiasLeo:
Configuration normalize now returns user model instead of internal config model
Added basic tests for normalize


lazan (approved):
I'll merge this via terminal

PR #63, opened 2022-07-14
Feature/1.5/patch domain multiplex send

lazan:


middleware/common/include/common/communication/device.h
```diff
@@ -350,23 +356,21 @@ namespace casual
...
-         {
-            // no-op - can't flush if we haven't got non-blocking
-         }
-
-         template< typename D>
-         static auto flush( D& device, traits::priority::tag< 1>) -> decltype( policy::non::blocking(device), void())
+         static auto flush( D& device, platform::size::type count, traits::priority::tag< 1>) -> decltype( policy::non::blocking(device), void())
          {
             // We don't want to handle any signals while we're flushing
             signal::thread::scope::Block block;
```
aliothe, line 362: Köas signalerna till processen upp tills end of scope o sedan är de tillgängliga?

lazan: New signals are blocked by the OS during the lifetime of the _scoped_,.  Could be "pending" (delivered from OS) signals ready to dispatch, these are also "blocked" by the _scoped_.

aliothe: Do they then get lost or are they delivered once the signal blocking is gone?

lazan: They do not get lost. Either the signal is "fired" during the block, and the OS will queue up the signal. Or, we got one or more signals ready for dispatch (ie we've got signals before but they are not consumed yet), the dispatch will also be "blocked" (no-op) during the _signal block_.  Hence, we don't need to "worry" about signal dispatch during the block. 
Of course can't anything be blocking during a _signal block scope_, or we might be stuck for ever.

PR #64, opened 2022-08-23
Feature/1.5/conversation send to terminated and user return code

OlofGullnas:
This PR fixes a bug where the tpsend and tprecv calls failed to set the tpurcode global variable. That is, they failed to report the "user return code" provided by a conversation subordinate service in its call to tpreturn. This affected both "normal" tpreturn calls with success and tpreturn calls that report TPFAIL. Also the somewhat esoteric case of the service calling tpreturn when not in control of the conversation.

The PR also includes some work to report an error on tpsend in subordinate when initiator has disconnected the conversation and in "initiator" when subordinate has called tpreturn (when not in control of conversation). This work was pending when the user return code problem was detected, and much of these changes are needed to return the user return code in all cases.

A number of unit test cases has been added for the conversational calls, but even more could be added.

One known bug remains. This occurs when a subordinate conversational service calls tpreturn when NOT in control of the conversation. If the tpreturn in this case reports TPFAIL, user return code and has a "data buffer", the result should be a TPEV_SVCERR in the initiator. This according to the XATMI spec. Currently a TPEV_SVCFAIL is reported, the same as if NO data buffer was passed in the tpreturn.

### DIFF
--- a/middleware/common/include/common/service/conversation/context.h
+++ b/middleware/common/include/common/service/conversation/context.h
@@ -52,6 +52,15 @@ namespace casual
             {
                using Flag = flag::service::conversation::send::Flag;
                using Flags = flag::service::conversation::send::Flags;
+               struct Result
+               {
+                  common::Flags< Event> event;
+                  long user = 0;
+                  CASUAL_LOG_SERIALIZE(
+                     CASUAL_SERIALIZE( event);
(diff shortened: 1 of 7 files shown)


OlofGullnas:
This PR is now obsolete. All changes in this PR are in PR #68 that has been merged.

Issue #67, opened 2022-09-05
Termination of server/subordinate handling a conversational service call is not reported to caller/initiator. 


OlofGullnas:

OlofGullnas:
PR #68 fixed this

PR #68, opened 2022-09-06
Feature/1.5/conversation service exit problem

OlofGullnas:
The strategy choosen to resolve the exiting process handling (issue #67 ) needs review! At least if the strategy choosen is reasonable. We may need to proceed with these changes (assuming that they are reasonable), while waiting for a better or cleaner solution. 

This PR includes the changes in PR #64.

This PR includes additional changes to handle issue #67.

The changes introduced in this PR affect error handling. In particular server processes that terminate unexpectedly while being busy with a conversational service. Could cause a hanging server (caller) to wait for a response that never will come. Could casuse serious problems for the system (lack of robustness).

The changes in PR #64 have to do with properly passing the user return code to the caller/initiator. Also error handling for services that behave badly, i.e. call tpreturn when not in control of a conversation or returning from service without calling tpreturn.

I am not certain that this is a good way to fix the issue, Do not know Casual internals well enough to say. But given my limited experience with Casual internals the changes seems reasonable. The key change is to use the correlation_id from a "lookup"  in the conversational connect. Instead of creating a new correlation_id. The other changes are to handle/recognize a possible service_reply message while doing tpsend or tprecv.

The changes seems to work in added unittests. The issue #67 changes will also be tested with the real application (with reintroduced application bug).

### DIFF
--- a/middleware/common/include/common/service/conversation/context.h
+++ b/middleware/common/include/common/service/conversation/context.h
@@ -52,6 +52,15 @@ namespace casual
             {
                using Flag = flag::service::conversation::send::Flag;
                using Flags = flag::service::conversation::send::Flags;
+               struct Result
+               {
+                  common::Flags< Event> event;
+                  long user = 0;
+                  CASUAL_LOG_SERIALIZE(
+                     CASUAL_SERIALIZE( event);
+                     CASUAL_SERIALIZE( user);
+                  )
+               };
             } // send
 
             namespace receive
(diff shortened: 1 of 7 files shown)


OlofGullnas:
Fix also worked with real applicarion. The initiator got a TPEV_SVCERR (as expected) when the server processing the subordinate service terminated.  The application rolled back and no hang occurred. Before fix the initiator would hang forever...

lazan (approved):

lazan:
I was going to "fix" the commit subjects, but I think it's best if you do it your self. 
The subject should start with   `<module> - <description of what is going to be done (fix/add...)>` note the lowercase to make it easier to find stuff in the log. The `<module>` might be _xatmi_ or _conversation_ ?

Either squash the WIP commits in to others, or have a proper subject. I suspect squash would be preferred, your call...

I find it easiest to use `git rebase -i  <sha1 of one earlier than the earliest you want to reorganise>`. You can always abort the rebase if anything goes wrong.

OlofGullnas:
Squashed a number of commits, updated commit comments and a comment. No changes to actual code.

PR #74, opened 2022-09-20
queue - recover queue message

ahhud:
recovery for queue messages with either commit or rollback


middleware/common/source/transaction/global.cpp
```diff
@@ -0,0 +1,61 @@
...
+         if( ! std::regex_match( string, std::regex{ "[a-f0-9]+"}))
+            code::raise::error( code::casual::invalid_argument, "invalid format on gtrid: ", string);
+
+         assertion( string.size() <= 64, "trid: ", string, " has larger gtrid size than 64");
+
+         common::transcode::hex::decode( string, gtrid.m_gtrid);
+         gtrid.m_size = string.size() / 2;
+
+         return in;
+
```
lazan, line 61: What if _input_ is not a multiple of 2 in hex? Let say `abc4345ff` ?

middleware/queue/include/queue/common/ipc/message.h
```diff
@@ -795,6 +796,59 @@ namespace casual
...
+               inline std::ostream& operator << ( std::ostream& out, recovery::Type value)
+               {
+                  switch( value)
+                  {
+                     using Enum = recovery::Type;
+                     case Enum::commit: return out << "commit";
+                     case Enum::rollback: return out << "rollback";
+                  }
+                  return out << "<unknown>";
+               }
```
lazan, line 816: Change to: 
`inline constexpr std::string_view description( Type value) noexcept` 
(then we can get rid of ostream and the includes all together).

middleware/queue/include/queue/common/ipc/message.h
```diff
@@ -795,6 +796,59 @@ namespace casual
                };
                
             } // remove
+            namespace recovery
+            {
+               enum class Type : int
+               {
+                  commit,
+                  rollback
+               };
```
lazan, line 805: Is it not more like _Phase_? It's what we want the transaction state to be when we're done, right? So more like _Phase_ or _Directive_?

middleware/queue/include/queue/group/queuebase.h
```diff
@@ -160,6 +160,12 @@ namespace casual
          //! @returns id:s of the messages that was removed
          std::vector< common::Uuid> remove( common::strong::queue::id queue, std::vector< common::Uuid> messages);
 
+         //! @returns gtrid:s of the messages that was commited
+         std::vector< common::transaction::global::ID> recovery_commit( common::strong::queue::id queue, std::vector< common::transaction::global::ID> gtrids);
+
+         //! @returns gtrid:s of the messages that was rollbacked
+         std::vector< common::transaction::global::ID> recovery_rollback( common::strong::queue::id queue, std::vector< common::transaction::global::ID> gtrids);
+
```
lazan, line 168: Feels intuitive.

middleware/queue/source/api/queue.cpp
```diff
@@ -276,7 +276,36 @@ namespace casual
...
+                        common::code::raise::error( queue::code::argument, "not possible to recover a remote queue: ", queuename);
 
+                     if( ! queue.process)
+                        common::code::raise::error( queue::code::no_queue);
+
+                     auto reply = common::communication::ipc::call( queue.process.ipc, request);
+
+                     common::log::line( verbose::log, "gtrids: ", reply.gtrids);
+
+                     return reply.gtrids;
```
lazan: I'm not sure if the move ctor will kick in. You might need to move reply.gtrids ?

middleware/queue/source/api/queue.cpp
```diff
@@ -276,7 +276,36 @@ namespace casual
...
+                     queue::Lookup lookup{ queuename};
+
+                     using Request = ipc::message::group::message::recovery::Request;
+                     Request request{ common::process::handle()};
+                     request.gtrids = gtrids;
+                     request.recovery_type = type;
+
+                     auto queue = lookup();
+
+                     if( queue.order > 0)
```
lazan: Use `queue.remote()` instead.

middleware/queue/source/group/handle.cpp
```diff
@@ -165,6 +165,27 @@ namespace casual
                      }
                      
                   } // remove
+
+                  namespace recovery
+                  {
+                     auto request( State& state)
+                     {
+                        return [&state]( const queue::ipc::message::group::message::recovery::Request& message)
+                        {
```
lazan, line 174: should not be `const` if you're intention is to move from the message.

middleware/queue/source/group/handle.cpp
```diff
@@ -165,6 +165,27 @@ namespace casual
...
+
+                           auto reply = common::message::reverse::type( message);
+
+                           if( message.recovery_type == queue::ipc::message::group::message::recovery::Type::commit)
+                              reply.gtrids = state.queuebase.recovery_commit( message.queue, std::move( message.gtrids));
+                           else
+                              reply.gtrids = state.queuebase.recovery_rollback( message.queue, std::move( message.gtrids));
+
+                           state.multiplex.send( message.process, reply);
+                        };
```
lazan, line 186: Looking nice!

middleware/queue/source/group/queuebase.cpp
```diff
@@ -571,6 +573,40 @@ namespace casual
...
+            const auto rollback2_affected = m_connection.affected();
+            m_statement.rollback3.execute( id());
+
+            return rollback1_affected == 0 && rollback2_affected == 0;
+         };
+
+         algorithm::container::trim( gtrids, algorithm::remove_if( gtrids, missing_message));
+
+         return gtrids;
+      }
```
lazan, line 609: Pretty easy to understand.

middleware/queue/source/manager/admin/cli.cpp
```diff
@@ -1138,7 +1140,77 @@ casual queue --restore <queue-name>)"
...
+                     auto option()
+                     {
+                        auto invoke = []( const std::string& queue, common::transaction::global::ID gtrid, std::vector< common::transaction::global::ID> gtrids)
+                        {
+                           gtrids.insert( std::begin( gtrids), std::move( gtrid));
+                           auto commited = queue::messages::recovery::commit( queue, gtrids);
+
+                           algorithm::for_each( commited, []( auto& gtrid)
+                           {
+                              std::cout << gtrid << '\n';
```
lazan: Use `common::log::line( std::cout, gtrid);`, so the `casual` serialization stuff kicks in... (or common::stream::write).

More to "follow" a "rule". always do this, to mitigate problems sometimes...

middleware/queue/source/manager/admin/cli.cpp
```diff
@@ -1138,7 +1140,77 @@ casual queue --restore <queue-name>)"
...
+                     auto option()
+                     {
+                        auto invoke = []( const std::string& queue, common::transaction::global::ID gtrid, std::vector< common::transaction::global::ID> gtrids)
+                        {
+                           gtrids.insert( std::begin( gtrids), std::move( gtrid));
+                           auto rollbacked = queue::messages::recovery::rollback( queue, gtrids);
+
+                           algorithm::for_each( rollbacked, []( auto& gtrid)
+                           {
+                              std::cout << gtrid << '\n';
```
lazan: Same here...

lazan (commented):
Looking good in general. I think you need to adress a few of my comments.

Also, I really don't understand why we "connect/group  queue (name) to gtrids"? Are there some technical issues that force it? 
As a user I would prefer just to commit/rollback a bunch of gtrids regardless of which queue the prepared messages are "located" at.

ahhud:
>

ahhud:
> Looking good in general. I think you need to adress a few of my comments.
> 
> Also, I really don't understand why we "connect/group queue (name) to gtrids"? Are there some technical issues that force it? As a user I would prefer just to commit/rollback a bunch of gtrids regardless of which queue the prepared messages are "located" at.

True! My main problem is to know which casual-queue-group to send recover to. 
To be able to be agnostic to queue I need to be able to ask which casual-queue-groups is involved in the transaction. And that I didn´t knew how to do. :-) Is there some way to do that?

lazan:
>  To be able to be agnostic to queue I need to be able to ask which casual-queue-groups is involved in the transaction. And that I didn´t knew how to do. :-) Is there some way to do that?

Send to all groups and just aggregate the replies? We got the _coordination abstraction_ for it, so it should be pretty easy.  

* queue-manager gets a service call for recovery
* create a recovery::Request and use message::coordinate to help with the _fan out_
* the service returns all gtrid that was _recovered_, hence the caller (cli, web) can use intersection/difference to let the user know which gtrid was altered.  If we want, we could return the associated queue-name and message-id for each affected gtrid (if it's easy), this could be something we could use to improve the cli over time, certainly a web front end could be richer with this information.  

I really do think it's a lot better if we can get rid of _queue-name grouping_ in the semantics. This affects the api and we'll be stuck with it for a while.

Note: If we use this semantics there is no "error" to try to recover an unknown gtrid, we just return the ones we've found and recovered. API:s like this is a lot better in my book.

ahhud:
> > To be able to be agnostic to queue I need to be able to ask which casual-queue-groups is involved in the transaction. And that I didn´t knew how to do. :-) Is there some way to do that?
> 
> Send to all groups and just aggregate the replies? We got the _coordination abstraction_ for it, so it should be pretty easy.
> 
> * queue-manager gets a service call for recovery
> * create a recovery::Request and use message::coordinate to help with the _fan out_
> * the service returns all gtrid that was _recovered_, hence the caller (cli, web) can use intersection/difference to let the user know which gtrid was altered.  If we want, we could return the associated queue-name and message-id for each affected gtrid (if it's easy), this could be something we could use to improve the cli over time, certainly a web front end could be richer with this information.
> 
> I really do think it's a lot better if we can get rid of _queue-name grouping_ in the semantics. This affects the api and we'll be stuck with it for a while.
> 
> Note: If we use this semantics there is no "error" to try to recover an unknown gtrid, we just return the ones we've found and recovered. API:s like this is a lot better in my book.

Sounds like a good plan. I have a go at it. :-)

middleware/queue/source/group/handle.cpp
```diff
@@ -165,6 +165,27 @@ namespace casual
...
+                     auto request( State& state)
+                     {
+                        return [&state]( queue::ipc::message::group::message::recovery::Request& message)
+                        {
+                           Trace trace{ "queue::handle::local::message::recovery::request"};
+                           log::line( verbose::log, "message: ", message);
+
+                           auto reply = common::message::reverse::type( message);
+
+                           if( message.directive == queue::ipc::message::group::message::recovery::Directive::commit)
```
lazan: nitpic: `if( message.directive == decltype( message.directive)::commit))`
 
Shorter and better for refactoring, and one can argue: more precise.

middleware/queue/source/manager/admin/cli.cpp
```diff
@@ -87,6 +89,22 @@ namespace casual
...
                }
 
+               std::vector< common::transaction::global::ID> recover( const std::vector< common::transaction::global::ID>& gtrids,
+                  ipc::message::group::message::recovery::Directive directive)
+               {
+                  serviceframework::service::protocol::binary::Call call;
+                  call << CASUAL_NAMED_VALUE( gtrids);
+                  call << CASUAL_NAMED_VALUE( directive);
+
+                  auto reply = call( manager::admin::service::name::recover);
```
lazan: Somewhat nitpicky: 
```
using Call = serviceframework::service::protocol::binary::Call;
return Call{}( manager::admin::service::name::recover, std::move( gtrids), std::move( directives)).extract<  std::vector< common::transaction::global::ID>>();
```

Replaces 8+ lines with 2.

middleware/queue/source/manager/admin/cli.cpp
```diff
@@ -1138,7 +1156,73 @@ casual queue --restore <queue-name>)"
...
+                           if( help)
+                              return { "<gtrid>"};
+
+                           return { "<value>"};
+                        };
+
+                        return argument::Option{
+                           std::move( invoke),
+                           complete,
+                           {  "--recover-transaction-commit"},
```
lazan: --recover-transaction-commit**s**.  cardinality are `1..*`, so it should be plural?

Or better: `--recover-transactions-commit`... 

Hmm, doesn't fell right regardless... Maybe something like: 

`casual transaction --recover-commit`  and `casual transaction --recover-rollback` is the best?

Language wise it's probably better with `casual transaction --commit-recover`, but then we loose the _grouping_ lexicographically, for the mind and auto-complete. So my vote is on the former.

middleware/queue/source/manager/admin/cli.cpp
```diff
@@ -1138,7 +1156,73 @@ casual queue --restore <queue-name>)"
...
+                           if( help)
+                              return { "<gtrid>"};
+
+                           return { "<value>"};
+                        };
+
+                        return argument::Option{
+                           std::move( invoke),
+                           complete,
+                           {  "--recover-transaction-rollback"},
```
lazan: Same here...

middleware/queue/source/manager/admin/server.cpp
```diff
@@ -179,6 +179,31 @@ namespace casual
...
+               });
+
+               return algorithm::accumulate( recover_futures, std::vector< common::transaction::global::ID>{}, []( auto result, auto& feature)
+               {
+                  auto predicate = [&result]( auto& value)
+                  {
+                     return ! algorithm::find( result, value);
+                  };
+
+                  algorithm::copy_if( future_get( feature).gtrids, std::back_inserter( result), predicate);
```
lazan, line 202: Nice, pretty easy to understand. clean code.

middleware/queue/unittest/source/test_queue.cpp
```diff
@@ -741,6 +746,126 @@ namespace casual
...
       }
 
+      namespace local
+      {
+         namespace
+         {
+            std::vector< common::transaction::global::ID> recover( const std::vector< common::transaction::global::ID>& gtrids,
+               ipc::message::group::message::recovery::Directive directive)
+            {
+               serviceframework::service::protocol::binary::Call call;
```
lazan: Same reduction in lines of code can be done here.

lazan (dismissed):
"approved", I do think you need to change the cli argument names. If so, you probably want to change the usage of _call_. 
If you argue that the current cli arguments are the right ones, you're free to merge. But subjective I think my proposal is better and more _casual_.

lazan (approved):

PR #75, opened 2022-09-22
transaction - list external resources from cli

AlbsAgren:
added cli option --list-external-resources: lists id, alias and pid of known external resources

### DIFF
--- a/middleware/transaction/source/manager/admin/cli.cpp
+++ b/middleware/transaction/source/manager/admin/cli.cpp
@@ -73,6 +73,17 @@ namespace casual
 
                   } // call
 
+                  namespace external
+                  {
+                     using base_type = admin::model::resource::external::Proxy;
+                     struct Proxy : base_type
+                     {
+                        Proxy() = default;
+                        Proxy( base_type base, std::string alias): base_type{ base}, alias{ std::move( alias)} { }
+
+                        std::string alias;
+                     };
+                  } // external
 
                   namespace format
                   {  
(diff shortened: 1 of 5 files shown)


test/unittest/source/test_transaction.cpp
```diff
@@ -372,6 +376,93 @@ namespace casual
...
+            -  connections: 
+               -  address: 127.0.0.1:7001
+)";
+
+            auto b = local::cli::domain( B);
+
+            auto a = local::cli::domain( A);
+            gateway::unittest::fetch::until( gateway::unittest::fetch::predicate::outbound::connected());
+
+            EXPECT_TRUE(administration::unittest::cli::command::execute( R"(casual transaction --list-external-resources --porcelain true)").string().empty());
```
lazan, line 452: nitpick: ws after `(`

middleware/transaction/source/manager/admin/cli.cpp
```diff
@@ -323,6 +348,35 @@ namespace casual
                         
                      } // resources
 
+                     namespace external
+                     {
+                        auto option()
+                        {
+                           return argument::Option{
+                              []()
```
lazan, line 356: nitpick: This, to me, feels like a strange format.

lazan (approved):
Looking good!

One thing. The unittest is as "far out" as we can come. I would prefer a unittest that use the model to verify that we have the correct state. 
We still need unittest for the cli though, so this might just be waste of time... 
If the logic was more complex and we needed more unittest, the model-way would be the right one, but for this cli-way only might be the best...

What do @ahhud and @albinsjolund think?

albinsjolund:
> Looking good!
> 
> One thing. The unittest is as "far out" as we can come. I would prefer a unittest that use the model to verify that we have the correct state. We still need unittest for the cli though, so this might just be waste of time... If the logic was more complex and we needed more unittest, the model-way would be the right one, but for this cli-way only might be the best...
> 
> What do @ahhud and @albinsjolund think?

I agree. I think we should look at cli testing and model testing as two different things, so cli testing should not replace testing the model, even if it could work sometimes for simple models :)

I was also wondering if we should have all cli tests in one place? I have only seen them in _middleware/administration/unittest/source/cli_. Or can't we access the transaction stuff from there @lazan ?

albinsjolund (approved):

middleware/transaction/source/manager/admin/cli.cpp
AlbsAgren: Is this preferable?
 ```
return argument::Option{ []()
    {
       ...
 ```
 I followed the convention used in the file but I can refactor all ocurrences if we think its an improvement.

PR #78, opened 2022-09-26
service - fix metrics to use route name

AlbsAgren:


middleware/gateway/source/group/inbound/state.cpp
```diff
@@ -37,7 +37,10 @@ namespace casual
                   message.pending = lookup.pending;
 
                   if( message.service.name != lookup.service.name)
+                  {
+                     message.service.requested = message.service.name;
                      message.service.name = lookup.service.name;
+                  }
```
lazan: You could do some exchange or swap instead of copies, right?

test/unittest/source/test_service.cpp
```diff
@@ -248,6 +250,41 @@ namespace casual
...
+         routes: [ a, b]
+)");
+
+         // Subscribe to the metric event
+         common::message::event::service::Calls event;
+         common::event::subscribe( common::process::handle(), { event.type()});
+
+         // Call service to generate metric
+         auto buffer = tpalloc( X_OCTET, nullptr, 128);
+         auto len = tptypes( buffer, nullptr, nullptr);
```
lazan: A scope around the variables (and usage of the variables). Mitigates unwanted usage and a good practice in general to keep scopes as small as possible.

lazan (changes requested):
Looking good, and easy to understand. The unittest was very concise and easy to understand.

A few pointers to fix though.

lazan (dismissed):
Nice!

ahhud (approved):

PR #83, opened 2022-10-03
common - add set container abstraction with contiguous memory

### LINKED ISSUE #80: generic container type that is always sorted
We need a container type that is always sorted and where operations on the container can't mess with the sorting.
I.e. appending and algorithms won't influence sorting.

lazan:
A container that has the following attributes:

* always sorted
* unique set of elements
* contiguous memory
* elements are immutable
* most of std::set API

fixes #80

### DIFF
--- a/middleware/common/include/casual/container/sorted/set.h
+++ b/middleware/common/include/casual/container/sorted/set.h
@@ -1,0 +1,178 @@
+//!
+//! Copyright (c) 2022, The casual project
+//!
+//! This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+//!
+
+#pragma once
+
+#include "common/traits.h"
+
+#include "common/algorithm/sorted.h"
+#include "common/algorithm/container.h"
+#include "common/compare.h"
+
+#include "common/serialize/macro.h"
+
+#include <vector>
+
+namespace casual
+{
+   namespace common::traits::is::allocator
+   {
+      //! traits to help deduce if type is allocator like.
+      //! if needed elsewhere, we move it to common/traits.h
+      template< typename Allocator, typename = void, typename = void>
+      struct like : std::false_type {};
+
+      template< typename Allocator>
+      struct like< Allocator,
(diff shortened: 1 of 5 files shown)


albinsjolund (approved):

lazan:
@ahhud You're the primary user of this container, at least right now. I think it would be good if you also did a review. @AlbsAgren You game?

AlbsAgren:
What is the main difference between this and std::set?

lazan:
> What is the main difference between this and std::set? 

`std::set` uses a node base memory "layout". Every element is allocated on the heap, by it self (a'la java). Gives a fem really good attributes, the most important is probably that insert and erase doesn't effect any iterators or references to other elements in the container.  A few big O algorithm improvements as a consequence of the _memory layout_., but in practice, with modern hardware, the allocation schema is way too slow and will dominate compare to a contiguous memory layout container, in general. But on paper it's pretty good :)
Also, it has to be implemented as a dubble linked list, hence no random access, and iteration involves chasing pointers -> added indirection for each step  -> cpu has no way to predict what to prefetch next -> wanted values not in L1 cache -> high chance of page fault -> fetch from main memory is orders of magnitudes more expensive -> SLOW
But if you need stable memory for references, this is probably a good choice.  

`casual::container::sorted::Set` uses contiguous memory layout (an internal `std::vector`). Always sorted (and unique) we can use binary search with `lower_bound` which has log N :ish characteristics -> size will not matter that much for search performance. Most likely slower for few elements compared to linear search, but in general a safer bet for all sizes. 
We will use this type for stuff that will be serialized in messages, hence a lot of iteration. Since the memory is contiguous the cpu will prefetch to L1 and next element will be "free" (regarding memory fetch). 

Hence, the differens is only for performance. And the fact that I personally hate reference semantics (except in extremely few places where a problem really "dictates" it)

AlbsAgren (approved):

ahhud (approved):

Issue #85, opened 2022-10-04
fix build environment sourcing compatibility with zsh and bash


lazan:
For some reason Apple, again, goes it's own way and don't care about the industry defacto standards... For better and certainly, for worse. They chosen zsh as the default shell, and make it unnecessary difficult to keep using bash (by design I presume). Hence, it's probably easier to have our few scripts compatible.

@albinsjolund You mentioned that you'd succeeded fixing the scripts?

PR #90, opened 2022-10-05
build - fix build environment sourcing compatibility with zsh and bash

albinsjolund:


lazan (approved):

lazan (changes requested):
Fix commit message with what you've done. I know it's pretty much the subject in other words, but practice makes perfect 😄 
Also, reference the issue in the commit. `fixes #85`  and NOT in the subject line. Every byte counts in the subject and it would be insane to pollute it with stuff that does not matter over time for the commit. The issue is of course interested in the relation during it's lifetime. 

Force push, of course.

lazan (approved):
Nice!

PR #92, opened 2022-10-13
Feature/1.6/fix creating unnecessary paths

AlbsAgren:
WIP

* Expanded the information that child- and grandchild processes may provide when connecting to the DM. Added a handler to the dm which responds to requests for information about the dm:s offspring.
* Fixed creating all of casuals directories whenever the casual cli was invoked.
* The TM now assigns external resources resource-id:s based on ipc rather than pid in to allow a process to have multiple external resources.

### DIFF
--- a/middleware/common/include/common/file.h
+++ b/middleware/common/include/common/file.h
@@ -140,9 +140,12 @@ namespace casual
          //! takes soft links into account when creating the path, if any.
          std::filesystem::path create( std::filesystem::path path);
 
+         //! creates all parent directories, but not the leaf
+         std::filesystem::path create_parent_path( std::filesystem::path path);
+
          namespace shared
          {
-            //! creates directory with group write.
+            //! creates directory with group rwx.
             std::filesystem::path create( std::filesystem::path path);
          } // shared
 
--- a/middleware/common/include/common/file.h
+++ b/middleware/common/include/common/file.h
@@ -140,6 +140,12 @@ namespace casual
          //! takes soft links into account when creating the path, if any.
          std::filesystem::path create( std::filesystem::path path);
 
+         namespace shared
+         {
+            //! creates directory with group write.
+            std::filesystem::path create( std::filesystem::path path);
+         } // shared
+
       } // directory
 
    } // common
--- a/middleware/common/include/common/message/transaction.h
+++ b/middleware/common/include/common/message/transaction.h
@@ -368,7 +368,7 @@ namespace casual
                         return involved;
                      }
                   } // involved
-               } // domain
+               } // external
             } // resource
          } // transaction
 
--- a/middleware/common/source/domain.cpp
+++ b/middleware/common/source/domain.cpp
@@ -116,6 +116,7 @@ namespace casual
             const Model model{ process::handle(), domain::identity()};
 
             auto& path = environment::domain::singleton::file();
+            directory::create( path.parent_path());
 
             if( std::filesystem::exists( path))
             {
--- a/middleware/common/source/domain.cpp
+++ b/middleware/common/source/domain.cpp
@@ -115,8 +115,7 @@ namespace casual
 
             const Model model{ process::handle(), domain::identity()};
 
-            auto& path = environment::domain::singleton::file();
-            directory::create( path.parent_path());
+            auto path = directory::create_parent_path( environment::domain::singleton::file());
 
             if( std::filesystem::exists( path))
             {
--- a/middleware/domain/include/domain/manager/admin/model.h
+++ b/middleware/domain/include/domain/manager/admin/model.h
@@ -169,6 +169,19 @@ namespace casual
             )
          };
 
+         struct Grandchild
+         {
+            common::process::Handle handle;
+            std::string alias;
+            std::filesystem::path path;
+
+            CASUAL_CONST_CORRECT_SERIALIZE(
+               CASUAL_SERIALIZE( handle);
+               CASUAL_SERIALIZE( alias);
+               CASUAL_SERIALIZE( path);
+            )
+         };
+
          namespace state
          {
             enum class Runlevel : int
@@ -200,6 +213,7 @@ namespace casual
             std::vector< model::Group> groups;
             std::vector< model::Executable> executables;
             std::vector< model::Server> servers;
+            std::vector< model::Grandchild> grandchildren;
 
             struct Tasks
             {
@@ -229,6 +243,7 @@ namespace casual
                CASUAL_SERIALIZE( groups);
                CASUAL_SERIALIZE( executables);
                CASUAL_SERIALIZE( servers);
+               CASUAL_SERIALIZE( grandchildren);
                CASUAL_SERIALIZE( tasks);
                CASUAL_SERIALIZE( event);
             })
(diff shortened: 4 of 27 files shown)


middleware/domain/source/manager/handle.cpp
```diff
@@ -908,6 +908,34 @@ namespace casual
...
+                        auto reply = common::message::reverse::type( message);
+                        algorithm::for_each( message.handles, [&state, &reply]( const auto& handle)
+                        {
+                           if( handle.pid)
+                           {
+                              if( auto server = state.server( handle.pid))
+                                 reply.processes.emplace_back( server->instance( handle.pid)->handle, server->alias, server->path);
+                              else if( auto executable = state.executable( handle.pid))
+                                 // we make sure to return a handle with pid only, since executables lack ipc
+                                 reply.processes.emplace_back( common::process::Handle{ handle.pid}, executable->alias, executable->path);
```
AlbsAgren, line 928: Bit unsure about this. But it makes some amount of sense that one should be able to get information about executables as well.

middleware/common/include/common/file.h
```diff
@@ -140,6 +140,12 @@ namespace casual
          //! takes soft links into account when creating the path, if any.
          std::filesystem::path create( std::filesystem::path path);
 
+         namespace shared
+         {
+            //! creates directory with group write.
+            std::filesystem::path create( std::filesystem::path path);
+         } // shared
```
lazan, line 150: Good name, I think.

lazan (approved):
My intention is to approve commit 66d4bdf. The GUI give me the notion that this is exactly what I'm doing. We'll see. I do not wan't to approve the whole PR...

middleware/common/include/common/message/transaction.h
```diff
@@ -368,7 +368,7 @@ namespace casual
                         return involved;
                      }
                   } // involved
-               } // domain
+               } // external
```
lazan, line 371: Good work updating spelling and stuff.

middleware/common/source/domain.cpp
```diff
@@ -116,6 +116,7 @@ namespace casual
             const Model model{ process::handle(), domain::identity()};
 
             auto& path = environment::domain::singleton::file();
+            directory::create( path.parent_path());
```
lazan: One liner? 
```
auto& path = directory::create( environment::domain::singleton::file());
```

Hence, `directory::create` need be able to understand what it is suppose to create.

middleware/domain/include/domain/manager/admin/model.h
```diff
@@ -229,6 +243,7 @@ namespace casual
                CASUAL_SERIALIZE( groups);
                CASUAL_SERIALIZE( executables);
                CASUAL_SERIALIZE( servers);
+               CASUAL_SERIALIZE( grandchildren);
```
lazan, line 246: I think this is a valid and good information to expose.

middleware/domain/include/domain/manager/state.h
```diff
@@ -299,8 +299,9 @@ namespace casual
          };
          std::string_view description( Runlevel value);
 
-      } // state
+         using Grandchild = common::message::domain::process::Information;
```
lazan, line 302: Nice.

middleware/domain/source/manager/handle.cpp
```diff
@@ -695,12 +695,12 @@ namespace casual
                   {
                      common::process::Handle pid( const State& state, strong::process::id pid)
                      {
-                        auto server = state.server( pid);
-                        
-                        if( server)
+                        if( auto server = state.server( pid))
```
lazan, line 698: 👍

middleware/domain/unittest/source/manager/test_manager.cpp
```diff
@@ -1526,5 +1526,133 @@ namespace casual
...
+         request.handles.push_back( process::handle());
+         // ...and the DM itself...
+         request.handles.push_back( domain.handle());
+         // ...as well as a non-existent pid.
+         request.handles.push_back( common::process::Handle{ common::strong::process::id{}});
+
+         auto processes = communication::ipc::call( communication::instance::outbound::domain::manager::device(), request).processes;
+
+         // The invalid pid should not be included in the response since it's not a real process
+         EXPECT_TRUE( processes.size() == 2);
```
lazan, line 1634: Probably good enough validation, but should you not verify that the range hold expected _processes_ ? Naively, you could use `algorithm::includes` and compare with a range of _pids_.

middleware/domain/unittest/source/manager/test_manager.cpp
```diff
@@ -1526,5 +1526,133 @@ namespace casual
```
lazan, line 1526: Good work. Easy to follow what is tested.

middleware/queue/source/group/handle.cpp
```diff
@@ -400,7 +400,7 @@ namespace casual
                            state.multiplex.send( message.process.ipc, reply);
                         };
                      }
-                  }
+                  } // prepare
```
lazan, line 403: 👍

middleware/queue/source/manager/handle.cpp
```diff
@@ -592,7 +593,10 @@ namespace casual
 
             auto spawn = []( auto& entity)
             {
-               entity.process.pid = common::process::spawn( state::entity::path( entity), {});
+               entity.process.pid = common::process::spawn(
+                  state::entity::path( entity),
+                  {},
+                  { instance::variable( instance::Information{ entity.configuration.alias})});
```
lazan, line 599: Hmm, we might wan't to _extend_ `process::spawn` to make it more strict what is "needed", instead "communicating" with environment variables... 

Either way, I don't think this commit/feature should improve that. Just a thought.

middleware/transaction/source/manager/admin/cli.cpp
```diff
@@ -352,23 +352,30 @@ namespace casual
...
-                              []()
+                           return argument::Option{ []()
                               {
-                                 auto state = call::state();
-
                                  auto transform = []( auto&& externals)
                                  {
-                                    return algorithm::accumulate( externals, std::vector< local::external::Proxy>{}, []( auto result, auto& external)
+                                    auto request = message::domain::process::information::Request{ process::handle()};
+                                    request.handles = algorithm::transform( externals, []( const auto& e) { return e.process;});
```
lazan, line 360: `e` ? vad betyder det?

lazan (dismissed):
Looking good. I thing you might want change a few things, but it's your call.

middleware/common/source/domain.cpp
AlbsAgren: I don't know if this is entirely trivial. For example, in `domain() / detail::casual / "singleton";` 'singleton' is a file and shouldnt be created as a dir by directory::create. But in `persistent() / "queue"` 'queue' is a directory and should be created.

A simple solution might be to have a create_parent_path function that only creates the parent directories but still returns the full path?

lazan: Yeah... you're right, it has to be explicit.

lazan (dismissed):

lazan:
@AlbsAgren. Can you "finalise" the commits? So, it's clear from the git log what has been done...  Force push it.

PR #93, opened 2022-10-21
queue - fix --list-remote

AlbsAgren:
A "minimum viable fix" for --list-remote, leaves the model as-is. It could be argued that we should be displaying more/different information (about the outbound, for example).


lazan (changes requested):
Good work. Fix the commit message. (Even if this is a trivial commit the log should describe what has been done).

Issue #96, opened 2022-10-24
core - during serve transaction commit -> resource error -> rollback


lazan:
[...]

PR #97, opened 2022-10-24
transaction - fix core when resources in distributed prepare (service) fails

lazan:
Before: When the prepare stage failed we tried to _pop transaction_, which rendered a core -> we didn't have any on "the stack", by design.

Now:

* We don't core :)
* If transaction is "bad" we propagate this on the regular _reply.code.result_ channel -> service call gives error
* Fixed a few unittest where we didn't explicitly called tx_rollback() after tx_commit() returned TX_ROLLBACK.

fixes issue #96

### DIFF
--- a/middleware/common/include/common/server/handle/service.h
+++ b/middleware/common/include/common/server/handle/service.h
@@ -90,9 +90,34 @@ namespace casual
 
          auto execute_reply = execute::scope( [&]()
          {
+            auto normalize_code = []( auto& reply) -> decltype(auto)
+            {
+               using Result = decltype( reply.code.result);
+               if( reply.code.result != Result::ok)
+                  return reply;
+
+               // if transaction state is _not good_ we need to indicate this on the
+               // normal "reply code channel".
+               switch( reply.transaction.state)
+               {
+                  // TODO: not totally sure about the exact correlation between 
+                  //    transaction.state -> reply.code.result. For now, we use "the worst".
+                  using Enum = decltype( reply.transaction.state);
+                  case Enum::active:
+                     break;
+                  case Enum::rollback:
+                  case Enum::timeout:
+                  case Enum::error:
+                     reply.code.result = Result::service_error;
+                     break;
+               }
+               return reply;
+            };
+
             // Send reply to caller.
             if( send_reply)
-               policy.reply( message.process.ipc, reply);
+               policy.reply( message.process.ipc, normalize_code( reply));
+               
          });
 
 
@@ -147,7 +172,7 @@ namespace casual
             execute_reply.release();
             execute_error_reply.release();
 
-            // reply.code.result is used to set outcomme in the 'ack'. We might want a _forward_ code?
+            // reply.code.result is used to set outcome in the 'ack'. We might want a _forward_ code?
             reply.code.result = code::xatmi::ok;
 
             return;
--- a/middleware/common/source/transaction/context.cpp
+++ b/middleware/common/source/transaction/context.cpp
@@ -23,6 +23,9 @@
 #include "common/message/transaction.h"
 #include "common/event/send.h"
 
+#include "casual/overloaded.h"
+#include "casual/assert.h"
+
 // std
 #include <map>
 #include <algorithm>
@@ -89,7 +92,7 @@ namespace casual
             return instance();
          }
 
-         bool Context::empty() const
+         bool Context::empty() const noexcept
          {
             return m_transactions.empty();
          }
@@ -124,8 +127,12 @@ namespace casual
                      request.alias = instance::alias();
                      request.resources = std::move( names);
 
-                     return common::environment::normalize( 
+                     auto result = common::environment::normalize( 
                         communication::ipc::call( communication::instance::outbound::transaction::manager::device(), request)).resources;
+
+                     log::line( verbose::log, "result: ", result);
+
+                     return result;
                   }
 
                } // resource
@@ -144,24 +151,26 @@ namespace casual
 
             auto [ named, unnamed] = common::algorithm::stable::partition( resources, []( auto& r){ return ! r.name.empty();});
 
-            auto names = algorithm::transform( named, 
-               []( auto& resource){ return resource.name;});
+            auto transform_name = []( auto& resource){ return resource.name;};
 
-            auto configuration = local::resource::configuration( std::move( names));
+            auto configuration = local::resource::configuration( algorithm::transform( named, transform_name));
 
-            auto transform_resource = [&configuration]( auto&& predicate)
+            auto transform_resource = [ &configuration]( auto&& predicate)
             {
-               return [&configuration, predicate = std::move( predicate)]( auto& resource)
+               return [ &configuration, predicate = std::move( predicate)]( auto& resource)
                {  
                   auto is_configuration = [&]( auto& configuration){ return predicate( resource, configuration);};
 
                   if( auto found = algorithm::find_if( configuration, is_configuration))
                   {
+                     // consume the resource configuration (make it impossible to reuse it...)
+                     auto value = algorithm::container::extract( configuration, std::begin( found));
+
                      return Resource{
                         resource,
-                        found->id,
-                        found->openinfo,
-                        found->closeinfo};
+                        value.id,
+                        std::move( value.openinfo),
+                        std::move( value.closeinfo)};
                   }
 
                   event::error::raise( code::casual::invalid_configuration, "missing configuration for linked named RM: ", resource, " - check domain configuration");
@@ -184,7 +193,8 @@ namespace casual
 
             // Open the resources...
             // TODO semantics: Not sure if we can do this, or if users has to call tx_open by them self...
-            open();
+            if( auto code = open(); code != code::tx::ok)
+               code::raise::error( code, "failed to open during configure");
 
          }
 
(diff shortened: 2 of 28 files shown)


AlbsAgren (dismissed):

albinsjolund (dismissed):

OlofGullnas:
Hm... 
I have not studied the code (yet), but the description of the changes make me suspicious! Might be the description that I misinterprete... 
You say you fixed some unittests that didn't explicitly call tx_rollback after tx_commit returned TX_ROLLBACK. 
This sounds wrong! I do NOT think there is a need/requirement to call tx_rollback in this situation. The description of the TX_ROLLBACK return code for TX_COMMIT say:
[TX_ROLLBACK]
The transaction could not commit and has been rolled back. In addition, if the
transaction_control characteristic is TX_CHAINED, a new transaction is started.

The first sentence implies (at least to me) that the transaction has been rolled back, and therefore there is no need to call tx_rollback.

Conceptually tx_commit and tx_rollback both guarantee that the current transaction is "terminated". A call to tx_commit tries to commit, but rolls back if this fails, while tx_rollback always does a rollback.  

Of course, if  TX_CHAINED is in effect (a new transaction is started when a transaction is "finished"), then there is a new transaction active after a TX_ROLLBAQCK return from tx_commit() that may need to be terminated properly...

lazan:
> Conceptually tx_commit and tx_rollback both guarantee that the current transaction is "terminated". A call to tx_commit tries to commit, but rolls back if this fails, while tx_rollback always does a rollback.

Nice catch. I'll verify it.

lazan:
You're totally right. I feel a bit embarrassed :) I'll fix it.

OlofGullnas:
And the bullet:

- If transaction is "bad" we propagate this on the regular reply.code.result channel -> service call gives error

If this is the case that Casual starts a transaction before invoking a service, e.g. "transaction: auto" and no transaction is passed in the call, then I can understand it.

If it is a failure to start a transaction before calling the service code, then an immediate SVCERROR is reasonable.

In the case that the transaction started properly, that the service code returns to casual with "success" and that Casual then tries to commit the active transaction but fails, then an SVCERROR is also reasonable. (The active transaction on return may be a transaction different from the one started by Casual if the service has called tx_commit/tx_begin while active...)

If the transaction for a a service with "auto" is rolled back by Casual because the service returned SVCFAIL, the the failure and any user return code and data should obviously be passed back to the caller.

In a scenario where the service starts a transaction on its own (tx_begin, i.e. Casual does not start any transaction before calling the service), then it is up to the service to decide what should be returned to its caller when a call to tx_commit fails with TX_ROLLBACK. In most cases probably "service failure".

There is the somewhat odd case of
- a service called without a transaction,
- casual does not start a transaction (not auto...)
- sevice starts a transaction
- service returns to casual with a transaction active!

I do not know if the "transaction demarcation" spec has anything relevant to say about this scenario. What should Casual do in this case?

If the service returned "failure" to Casual, then the obvious choice is rollback of the transaction.

But what if "success" is returned? It can be argued that the service misbehaved and did not decide what to do with the transaction it (explicitly) started. This implies that Casual should do a rollback in this scenario ("presumed rollback"), but it can also be argued that a tx_commit should be done in this case. My first instinct is that tx_rollback (and SVCERROR) is the correct choice, as the application "misbeaved".

The main argument against this is maybe to simplify the application code when TX_CHAINED is in effect. In that case every tx_commit or tx_rollback will implicitly do a tx_begin, which implies that a transaction is active if a  tpreturn is done after a tx_commit. In this case it is not reasonable to do a rollback and return SVCERROR. This would loose the return data provided in the tpreturn (with success).

Requiring the application to call tx_set_transaction_control(TX_UNCHAINED) before its "last" call to tx_commit/tx_rollback is somewhat doubtful. May lead to convoluted application logic as the code needs to know before the commit (or rollback) if a new transaction is needed (or a tx_set_transaction_control(TX_UNCHAINED) + tx_rollback before the tpreturn). I do not know if using TX_CHAINED is common. I have never used it.

OlofGullnas:
And on the tx_rollback after a tx_commit (that you expected to fail in the test). Perhaps you should leave it in the unit test code but check that it returns TX_PROTOCOL_ERROR :-)

middleware/common/include/common/server/handle/service.h
```diff
@@ -90,9 +90,34 @@ namespace casual
...
          {
+            auto normalize_code = []( auto& reply) -> decltype(auto)
+            {
+               using Result = decltype( reply.code.result);
+               if( reply.code.result != Result::ok)
+                  return reply;
+
+               // if transaction state is _not good_ we need to indicate this on the
+               // normal "reply code channel".
+               switch( reply.transaction.state)
```
OlofGullnas, line 101: Not 100% sure about exact circumstances when this is executed. And what the "context" is. Looking at this "locally" it seems to be all cases where a service returns with reply.code.result == Result::ok.
Is there always a transaction state?
What if a service is called without a transaction? Not all services have transactions active when called. A service may start its own transaction(s), or may not use transactions at all. Caller of a service may specify NOTRAN (or has not started a transaction), and if the called service is not "auto"...
As I understand things in the XATMI world, if a service is called with a transaction, then that transaction should be put in "rollback only" if the called service or any service it calls with the same transaction returns an error (SVCFAIL or SVCERROR). 
A failure to start a transaction in a service that is auto (or any other case where a transaction is started by Casual on the services "behalf") shoulöd probably cause SVCERROR. A failure by casual to commit an auto transaction should also cause a SVCERROR return. There is no way to do better in that case. 

Perhaps the transaction policy handles all this...

lazan: Well, the actual impact, or rather what we are trying to achieve, is if the resource interaction fails we need to propagate this to caller. If xa_start fails, the (user) service code is not even invoked. From the caller perspective, the service failed, there's no question about that, right? I don't see how one could argue otherwise... This is stuff that happens outside user control. Either before service invokation, or after.

lazan:
> And on the tx_rollback after a tx_commit (that you expected to fail in the test). Perhaps you should leave it in the unit test code but check that it returns TX_PROTOCOL_ERROR :-)

Yep! 😄

middleware/common/include/common/server/handle/service.h
OlofGullnas: Sure, if e.g. xa_start fails during "service invocation" (before user code is called) it is correct to return TPESVCERR. Also if something fails at the end (xa_end...). TPESVCERR is the correct value to return for errors that occur during service invocation and service return. (There is also an error TPETRAN described for tpcall that is for tpcall without TPNOTRAN to a service that does not support transactions. This is also "kind of" an error during invocation...)
   
I was worried about the case of a service invoked with tpcall(..., TPNOTRAN) for a service that has for example "transaction: join". What is reply.transaction.state in that case? I assume there is no "transaction" in that case and it is not immediately obvious that the value is == "Enum::active". It MAY be initialized to that even without transaction... I investigated and it seems as if it is...  (at common/include/common/message/service.h line 151). So the code should work...

OlofGullnas: The other case I was thinking about was transactions started during service invocation ("transaction: auto" and no transaction passed from caller). The reasonable behaviour in that case is to commit the current transaction if the service returns TPSUCCESS, rollback if the service returns TPFAIL and TPESVCERR if an error occurs during commit (or rollback?).

I assume this is handled by execute_transaction().

The current transaction on tpreturn might be different from the one started during invocation if the service has called tx_commit & tx_begin while executing. This is probably "legal" if the transaction was started on service invocation... This is an unlikely case as a service with "auto" probably is coded with the expectation that it is invoked with a transaction active, and then it is not allowed to commit or rollback. It can try :-)

I wonder what happens if the (auto) service has called tx_commit or tx_rollback before tpreturn. This is unlikely in, but I have not investigated.

lazan:
We need to re-review this again. The latest commit differs a lot from the first time...
@albinsjolund @AlbsAgren @ahhud

ahhud (approved):

OlofGullnas:
@lazan 
Noticed that a lot had changed, 😄
Is the description of the pull request still correct? First two bullets are likely correct,..
Is the description reasonably "complete"?

Is there a recommended order to review the changes. I started from the "top" file in "files changed" and get the feeling that this is not the smartest thing to do.

middleware/common/source/transaction/context.cpp
```diff
@@ -555,7 +563,7 @@ namespace casual
 
             log::line( log::category::transaction, "transactions: ", transactions);
 
-            const auto process = process::handle();
+            //const auto process = process::handle();
```
albinsjolund, line 566: If not used anymore I think it should be removed and not just commented out.

middleware/common/source/transaction/context.cpp
```diff
@@ -581,93 +589,85 @@ namespace casual
...
-                  log::line( log::category::error, exception::capture(), " failed to rollback transaction: ", transaction.trid);
-               }
+               return local::log::code( Context::rollback( transaction), "failed to rollback transaction: ", transaction.trid);
             };
 
-            auto trans_commit_rollback = [&]( const Transaction& transaction)
+            auto invoke_commit_rollback = [ this, invoke_rollback, commit]( const Transaction& transaction)
             {
                if( commit && transaction.state == Transaction::State::active)
+                  //return local::log::code( Context::commit( transaction), "failed to commit transaction: ", transaction.trid);
```
albinsjolund, line 600: Same here, remove comment if not used.

albinsjolund (approved):

lazan:
> @lazan Noticed that a lot had changed, 😄 Is the description of the pull request still correct? First two bullets are likely correct,.. Is the description reasonably "complete"?
> 
> Is there a recommended order to review the changes. I started from the "top" file in "files changed" and get the feeling that this is not the smartest thing to do.

Well this is not a good PR, in a review sense... But I did struggle to shop it up in some logical grouping of commits.
The PR description did not update based on the commit message. In general, the commit messages are what's important. And the "PR description" is just _stuff regarding the PR_, not necessary what's have changed.

I tend to review commit by commit. Otherwise, top to bottom. And, I tend to ignore whitespace, do reduce change...

middleware/common/source/transaction/context.cpp
lazan: dammit, I totally missed this comment...

middleware/common/source/transaction/context.cpp
lazan: dammit, I totally missed this comment...

lazan: This also looks suspicious... I'll check it out...

OlofGullnas:
I had missed the commit comment...

lazan:
Rebased and merge locally, then pushed... GitHub did not pick up on it... don't know why

PR #107, opened 2022-11-10
gateway - immediately send error reply on failed resource request

### LINKED ISSUE #98: casual-queue-forward occasionally hangs on shutdown
Potentially caused by the TM not sending an error reply to the forward after it fails to send a rollback_request to an outbound-grup. This would cause the forward to be stuck with an instance it cannot get rid of even though the transaction has been rolled back elsewhere.

AlbsAgren:
Before:
We took no immediate action when failing to correlate a descriptor with a connection when sending resource requests.
This would occasionally lead to situations where casual-queue-forward would hang waiting on a commit/rollback reply.

Now:
We confirm that the message was successfully sent and immediately invoke the error-callback if not.

Fixes #98

### DIFF
--- a/middleware/administration/unittest/source/cli/test_transaction.cpp
+++ b/middleware/administration/unittest/source/cli/test_transaction.cpp
@@ -1,0 +1,106 @@
+//!
+//! Copyright (c) 2022, The casual project
+//!
+//! This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+//!
+
+#include "common/unittest.h"
+
+#include "domain/unittest/manager.h"
+
+#include "administration/unittest/cli/command.h"
+
+#include "gateway/unittest/utility.h"
+
+#include "casual/tx.h"
+#include "casual/xatmi.h"
+
+namespace casual
+{
+   using namespace common;
+
+   namespace administration
+   {
+      namespace local
+      {
+         namespace
+         {
+            namespace cli
+            {
+               constexpr auto base = R"(
+domain:
+   servers:
+      - path: "${CASUAL_MAKE_SOURCE_ROOT}/middleware/service/bin/casual-service-manager"
+      - path: "${CASUAL_MAKE_SOURCE_ROOT}/middleware/transaction/bin/casual-transaction-manager"
+      - path: "${CASUAL_MAKE_SOURCE_ROOT}/middleware/gateway/bin/casual-gateway-manager"
+)";
+
+               template< typename... C>
+               auto domain( C&&... configurations)
+               {
+                  return casual::domain::unittest::manager( base, std::forward< C>( configurations)...);
+               }
+
+               auto call( std::string_view service)
+               {
+                  auto buffer = tpalloc( X_OCTET, nullptr, 128);
+                  common::unittest::random::range( range::make( buffer, 128));
+                  auto len = tptypes( buffer, nullptr, nullptr);
+
+                  tpcall( service.data(), buffer, 128, &buffer, &len, 0);
+                  EXPECT_TRUE( tperrno == 0) << "tperrno: " << tperrnostring( tperrno);
+
+                  return memory::guard( buffer, &tpfree);
+               };
+
+            } // cli
+         } // <unnamed>
+      } // local
+
+      TEST( cli_transaction, list_external_resources__expect_outbound_gateways_to_be_listed)
+      {
+         common::unittest::Trace trace;
+
+         auto b = local::cli::domain( R"(
+domain: 
+   name: B
+
+   servers:
+      - path: "${CASUAL_MAKE_SOURCE_ROOT}/middleware/example/server/bin/casual-example-server"
+
+   gateway:
+      inbound:
+         groups:
+            -  connections: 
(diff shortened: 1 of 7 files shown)


middleware/administration/unittest/source/cli/test_transaction.cpp
```diff
@@ -0,0 +1,106 @@
...
+         ASSERT_TRUE( tx_commit() == TX_OK);
+
+         std::string output = administration::unittest::cli::command::execute( R"(casual transaction --list-external-resources --porcelain true | cut -d '|' -f 2)").string();
+
+         constexpr auto expected = R"(outbound_B
+)";
+         EXPECT_TRUE( output == expected) << "output:  " << output << "expected: " << expected;
+      }
+
+   } // administration
```
lazan, line 105: Nice testcase. Easy to follow. Good to actually test cli.

middleware/domain/source/manager/task/create.cpp
```diff
@@ -562,7 +562,7 @@ namespace casual
                }
             };
 
-            return manager::Task{ "remove alieases", scale::local::group::task( std::move( groups), std::move( done_callback)),
+            return manager::Task{ "remove aliases", scale::local::group::task( std::move( groups), std::move( done_callback)),
```
lazan, line 565: Nice!

middleware/gateway/source/group/outbound/handle.cpp
```diff
@@ -114,10 +114,12 @@ namespace casual
                                  state.multiplex.send( destination.ipc, reply);
                               });
 
-                              tcp::send( state, descriptor, message);
+                              // if we fail to send we want to invoke the error callback immediately
+                              if( ! tcp::send( state, descriptor, message).valid())
+                                 state.route.consume( message.correlation).error();
```
lazan, line 119: Should this be done generically by `tcp::send`. Naively, this could happen in other "flows"? Whats your thoughts?

middleware/gateway/unittest/source/test_manager.cpp
```diff
@@ -1271,6 +1271,85 @@ namespace casual
...
+         {
+            auto reply = common::communication::ipc::receive< common::message::transaction::commit::Reply>();
+
+            EXPECT_TRUE( reply.state == decltype( reply.state)::fail) << CASUAL_NAMED_VALUE( reply);
+            EXPECT_TRUE( reply.trid == transaction::context().current().trid);
+         }
+
+         EXPECT_EQ( transaction::context().commit(), code::tx::ok);
+      }
+
```
lazan, line 1356: Nice testcase, easy to follows.

middleware/gateway/unittest/source/test_manager.cpp
```diff
@@ -1271,6 +1271,85 @@ namespace casual
...
+         }
+
+         // we 'lose' our connection to b
+         {
+            common::signal::send( inbound.pid, code::signal::terminate);
+         }
+
+         // we try to commit our transaction
+         {
```
lazan, line 1339: There is a timing issue here. There is no guarantee when signals are delivered. You should check that the _outbound_ is non-existent, before committing.
Hence, keep a process, as your inbound, and then _fetch::until_ you can't find it in state

lazan (changes requested):
Very good, I would like you to fix the timing issue in one of the test-cases. Hence, not add instability :)

middleware/gateway/unittest/source/test_manager.cpp
```diff
@@ -1271,6 +1271,86 @@ namespace casual
...
+
+            EXPECT_TRUE( result.buffer.memory == payload.memory);
+
+         }
+
+         // b shuts down
+         {
+            communication::device::blocking::send( b.handle().ipc, common::message::shutdown::Request{});
+            unittest::fetch::until( unittest::fetch::predicate::outbound::connected( 0));
+         }
```
lazan, line 1336: It's still a timing issue. `gateway-manager` in B will start `b` as soon as it can. `fetch::until` might miss the _down time_. I'm guessing that it's very unlikely, but I'v been more surprised for less... I think you need to make sure that `b` is not B's inbound any more. You might need to _switch/activate to B_ , check, and then switch back to A.

Hence, for this to happen the predicate for `fetch::until` will never be true, and an error will be reported after a "while".  It might be very unlikely... We might roll with your current code. What's your take on it?

AlbsAgren: I figured (and observed) that we might miss the down time if we just sent a terminate signal to the inbound since b's dm will just restart it. But i thought that stopping the entire b domain would solve it, is that not the case? I might also argue that stopping the entire domain has the further 'advantage' of being more similar to the scenario where the issue was discovered (i.e. a pod with the proxy domain dying).

lazan, line 1336: You could do a: `unittest::sink( std::move( b));` to shutdown `B`. I approve, it's up to you.

lazan (dismissed):
Good work.

lazan (approved):

PR #109, opened 2022-11-15
queue - add browse-peek to queue api

lazan:
Add _browsing_ functionality to the queue api. User provided callback gets invoked with a message until the queue is empty or the callback returns false.

Corresponding C-api added as well.

We could expand on this in the future and have the same semantics for browse-dequeue. It might be a better api for the user.

### DIFF
--- a/middleware/queue/source/api/queue.cpp
+++ b/middleware/queue/source/api/queue.cpp
@@ -518,7 +518,7 @@ namespace casual
                   common::code::raise::error( queue::code::argument, "not possible to peek a remote queue: ", queuename);
 
                if( ! queue.process)
-                  common::code::raise::error( queue::code::no_queue);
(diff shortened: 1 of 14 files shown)


middleware/queue/source/api/queue.cpp
```diff
@@ -518,7 +518,7 @@ namespace casual
                   common::code::raise::error( queue::code::argument, "not possible to peek a remote queue: ", queuename);
 
                if( ! queue.process)
-                  common::code::raise::error( queue::code::no_queue);
+                  common::code::raise::error( queue::code::no_queue, "failed to lookup up: ", lookup.name());
```
AlbsAgren, line 521: Should be "failed to lookup queue: " I think

lazan:
Main reviewer: @albinsjolund

albinsjolund (approved):

Issue #106, opened 2022-11-09
casual should log which queuename a failed operation i made on


ahhud:
To be more user friendly casual should log the name of the queue that can´t be operated on.

i.e
2022-10-28T08:21:23.829790+02:00|xxx|yyy|zzz|ppp|xxx||||error|failed to lookup queue - action: send error reply

which is to generic

Issue #105, opened 2022-11-09
casual-domain-manager should log name of server who exists


ahhud:
In order to be more user friendly casual-domain-manager should log name of exited server isof only pid, status and reason.

PR #113, opened 2022-12-01
Feature/1.6/better logging on process exit

AlbsAgren:
Addresses #105 & #106


middleware/domain/source/manager/handle.cpp
```diff
@@ -570,10 +570,22 @@ namespace casual
                            // We don't want to handle any signals in this task
                            signal::thread::scope::Block block;
 
+                           auto get_alias = [ &state]( strong::process::id pid) -> std::string
+                           {
```
lazan, line 574: Could be std::string_view, but it doesn't matter I think

lazan (approved):
Nice!

PR #114, opened 2022-12-06
Feature/1.6/ipc/multiplex reconnect

lazan:
I think it's best to review commit by commit. I've tried to keep stuff together as best I can... 

It might be good to ignore ws change in some commit.

_"main reviewer:"_  @ahhud


middleware/common/include/common/communication/ipc/send.h
```diff
@@ -157,12 +158,67 @@ namespace casual
...
+            if( auto found = algorithm::find( m_destinations, device.connector().process().ipc))
+            {
+               if( found->push( std::move( message)).send( *m_directive))
+                  m_destinations.erase( std::begin( found));
+
+               return result;
+            }
+            else
+            {
+               while( true)
```
ahhud, line 186: There is not a risk for a never ending loop here?

.gitignore
```diff
@@ -11,52 +9,38 @@
...
 
-
 /Default/
 /Release/
 
-#
-# Tag files produced by Global
-#
+# Tag files produced by Global (what is this?)
 GPATH
```
OlofGullnas, line 36: Googlade. Fick up:
GNU Global is a source code tagging system that works the same way across diverse environments, such as Emacs editor, Vi editor, Less viewer, Bash shell, ...

Detta stämmer också med min minnesbild. Tyckte mig känna igen GTAGS med flera i ett Emacs context.

middleware/common/include/common/communication/ipc/send.h
lazan: It's to keep going if we've get multiple interups... (the reconnect throws if fail) The whole exception paradigm does not work well when the "errors" are expected... I'm going to change at least the whole _device_ API to communicate with _std::expected_ instead. Hast to be a lot better, since a lot of the "errors" are indeed expected...

.gitignore
lazan: Thanks

lazan:
@ahhud Any reservations? I think the last commit fixed the "random shutdown instance lookup deadlock" on unittests.

PR #118, opened 2022-12-12
forward - fix forwards dying when scaling aliases

AlbsAgren:
* We now restart the flow when receiving dequeue::forget::Replys, giving us a chance to start new forward instaces if we have to few.
* When complying to configuration changes, we only cancel as many forward instances as we have to, in order to cause less disruption to the "flow".

### DIFF
--- a/middleware/queue/include/queue/forward/state.h
+++ b/middleware/queue/include/queue/forward/state.h
@@ -101,6 +101,13 @@ namespace casual
                   return configured - running;
                }
 
+               inline platform::size::type surplus() const
+               {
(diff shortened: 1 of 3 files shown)


middleware/queue/include/queue/forward/state.h
```diff
@@ -101,6 +101,13 @@ namespace casual
                   return configured - running;
                }
 
+               inline platform::size::type surplus() const
```
lazan, line 104: Nice name.

lazan (approved):

PR #119, opened 2022-12-15
never call server_done unless server_init has been called

aliothe:


OlofGullnas:
What was the original problem?

I might be missing something, but:

If I understand this  change, it avoids calling tpsvrdone (i.e. argument.server_done) if no init routine (given by argument.server_init) was specified (and obviously not called). Having a "done" routine" but no "init" is a bit odd...

What if the init routine is unsuccessful or throws an exception? The code seems to say that a return value of -1 from the init routine is treated as init routine failure, and done.release() is called to avoid calling the "done" routine. All other return values from init are treated as success and the "done" will be called. Should the test for init routine failure be return value != 0?

What about an exception from the init routine? An exception should probably end up in a server exit (and it probably does), but should the argument.server_done routine be called when the stack is unwound in this case?

Is there some magic in common::invoke(argument.server_init, ...) that catches exceptions? (I have not investigated if there is!)

aliothe:
> What was the original problem?
> 
> I might be missing something, but:
> 
> If I understand this change, it avoids calling tpsvrdone (i.e. argument.server_done) if no init routine (given by argument.server_init) was specified (and obviously not called). Having a "done" routine" but no "init" is a bit odd...
> 
> What if the init routine is unsuccessful or throws an exception? The code seems to say that a return value of -1 from the init routine is treated as init routine failure, and done.release() is called to avoid calling the "done" routine. All other return values from init are treated as success and the "done" will be called. Should the test for init routine failure be return value != 0?
> 
> What about an exception from the init routine? An exception should probably end up in a server exit (and it probably does), but should the argument.server_done routine be called when the stack is unwound in this case?
> 
> Is there some magic in common::invoke(argument.server_init, ...) that catches exceptions? (I have not investigated if there is!)

Done without init is indeed a bit odd but in the case that happens, done should not be called.
That was the case that should be fixed.
Regarding the other things, that's nothing that I've checked nor added.

lazan (changes requested):
Fix the commit message to something that make sense in a couple of years.

PR #121, opened 2022-12-19
tpsrvdone should never be called if tpsrvinit has not been called

aliothe:


lazan (approved):
I'll merge this locally.

lazan:
Merged locally.

PR #123, opened 2022-12-22
Feature/1.6/build time visibility

lazan:


middleware/configuration/documentation/domain.general.operation.md
```diff
@@ -149,7 +149,6 @@ domain:
         timeout:
```
AlbsAgren, line 154: Github wont let me comment on the right lines, but on 53 & 132 discoverable is listed as a property, but it's now called 'visibility', right?

AlbsAgren (dismissed):

middleware/configuration/documentation/domain.general.operation.md
lazan: correct, good catch!

lazan: I'll fix that. But you're ok with the rest?

AlbsAgren: Yeah, looks good to me

PR #127, opened 2023-01-10
queue: use blocking semantic for 'transaction involve send'

lazan:
This patch make sure that the transaction involvement message to TM is sent with blocking semantics, to ensure that TM knows about the queue-group association with the transaction BEFORE the actual transaction commit phase (with prepare and so on).

### DIFF
--- a/middleware/queue/source/group/handle.cpp
+++ b/middleware/queue/source/group/handle.cpp
@@ -43,6 +43,7 @@ namespace casual
                {
                   namespace transaction
                   {
+                     //! blocking "involve" send to TM
                      template< typename M>
                      void involved( State& state, M& message)
                      {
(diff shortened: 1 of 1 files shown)


ahhud (approved):

OlofGullnas:
@lazan I took a quick look at this. As I understand it you changed from using state.multuiplex.send() to  communicatiojn::device::blocking::send().

I must admit that I have not studied the exact semantics of state.multiplex.send(), but:
If multiplex.send() preserves order, and all sends to the TM in the process use multiplex.send (the later part is perhaps not the case...), is there really a difference? Does completion of a blocking send gurantee that the message has been seen and processed(!) by the destination (TM)? In general I think the answer is no... Using the blocking send bypasses any queueing in the multiplex.send() abstraction, but there is still a potential buffering in the OS kernel or underlyng communication mechanism.

Are there race conditions/timing cases where other processes (and the "system") relies on the TM knowing about the "involvment"? Is there a need for an "ack" from the TM? An "ack" would guarantee that the TM has received and processed the "involved" message.

It is possible that the TM use of a single input "ipc" for everything (all inputs) guarantees "correct" processing order if/when a blocking send is used, Any issues with multi domain configurations?

lazan:
Well, if the message is actually delivered to the callees "ipc device", then the order is guaranteed, as I understand it. Before the patch, if the message could not be sent, `state.mutlplex.send` will just "cache" the message to be sent later. That is, we multiplex on all destination `state.mutlplex.send` has messages to be sent to, and try to send them when OS says it is possible (this does not work flawless, but that is another issue). 

This possible "severe" delay made it possible for the caller to trigger the desitributed commit/rollback and the local TM is invoked BEFORE the casual queue "involved" message. Hence the bug.

 I'm not sure if it is possible to get "out of order" for domain sockets if A -> X THEN B -> X, from different processes. I don't think so, but I'm not sure.
I'm sure that this is impossible for _named pipes/FIFO_, but not totally sure for domain-sockets. 

So, there could be a glitch still.

PR #130, opened 2023-01-16
configuration - add set operations union, intersection, difference

albinsjolund:
Added options to performe set operations on configuration For example:
casual domain --configuration-get | casual configuration --union new_config.yaml


middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
```diff
@@ -0,0 +1,145 @@
...
+)");
+
+      auto rhs = common::unittest::file::temporary::content(".yaml", R"(
+domain:
+   servers:
+      - alias: Server2
+      - alias: Server3
+)");
+
+      auto command = "casual domain --configuration-get | casual configuration --union " + rhs.string();
```
lazan, line 78: What if:  `casual domain --configuration-get json | ... `?

middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+      auto union_output = administration::unittest::cli::command::execute( command).string();
+
+      // Expect to find all 3 servers
+      EXPECT_TRUE( union_output.find("Server1") != std::string::npos) <<  union_output;
+      EXPECT_TRUE( union_output.find("Server2") != std::string::npos) <<  union_output;
+      EXPECT_TRUE( union_output.find("Server3") != std::string::npos) <<  union_output;
```
lazan, line 85: We should use something else than std::string index based "algorithms". Index based are more error prune than iterator based (at least as I see it), especially if "mixed" throughout the code base. 
I don't think this specific case is that bad though, it's more a principle...

middleware/configuration/unittest/source/test_model.cpp
```diff
@@ -454,6 +454,237 @@ namespace casual
...
+        path: a
+      - alias: c
+        path: c
+)";
+
+         auto A = local::configuration( configuration);
+         auto B = local::configuration( remove);
+         auto result = local::configuration( expected);
+
+         set_difference( A, std::move( B));
```
lazan: I'm not sure if I like that set_xxx are mutating... Feels wrong to update A in this case... To me, it would be more _functional_ to have `A = set_difference( std::move( A), std::move( B));` Or, if one need to keep originals: `auto result = set:difference( A, B);`

middleware/configuration/unittest/source/test_model.cpp
```diff
@@ -454,6 +454,237 @@ namespace casual
...
+)";
+
+         auto A = local::configuration( configuration);
+         auto B = local::configuration( add);
+         auto result = local::configuration( expected);
+
+         set_intersection( A, std::move( B));
+
+         EXPECT_TRUE( A == result) << CASUAL_NAMED_VALUE( A) << "\n" << CASUAL_NAMED_VALUE( result);
+      }
```
lazan, line 686: Nice set of test cases. Feels like we've got good coverage.

middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
albinsjolund: I have only tested that manually and it works, but I will add a testcase for that too.

middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
albinsjolund: Do you mean like this? `union_output.find( "Server1") < union_output.length()`
Or is it `find` I should not use? What other options do I have?

middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
lazan: How can it work?

middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
lazan: More like `EXPECT_TRUE( algoritm::find( union_output, "Server1"))`... Not sure if it will work with string literal, you might have to use string_view. `"Server1"sv`

Or, keep it as if.

middleware/administration/unittest/source/cli/configuration/test_set_operations.cpp
albinsjolund: I don't know. Json works but not xml

lazan (approved):

PR #136, opened 2023-02-01
Feature/1.6/ngins plugin improvments main

ahhud:


middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+            template <typename T>
+            auto copy( T value)
+            {
+               auto result = static_cast< char*>( malloc( value.size() + 1));
+               std::copy( std::begin( value), std::end( value), result);
+               result[ value.size()] = '\0';
+               return result;
+            }
```
lazan, line 83: This is only for copying/mallocing "string-like" stuff? If so, should it be:
`char* copy( std::string_view value)` ? to get more help from compiler?

middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+            auto allocate()
+            {
+               return static_cast<T*>( malloc( sizeof(T)));
+            }
+
+            template <typename T>
+            void deallocate( T* ptr)
+            {
+               free( ptr);
+               ptr = nullptr;
```
lazan: `ptr` is by value, so setting it to nullptr does not do anything to outside the function.

middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+         }
+         namespace memory
+         {
+            template <typename T>
+            auto allocate()
+            {
+               return static_cast<T*>( malloc( sizeof(T)));
+            }
```
lazan: Seems a bit dangerous? :) I would like to see some sort of restriction of which types can be _malloced_, Such as:
```
 template <typename T>
 auto allocate() -> std::enable_if_t< std:: is_trivially_copyable_v< T>, T*>
{
   ... 
}

File: middleware/http/source/inbound/c/api.cpp [hunk 4]
```diff
@@ -0,0 +1,254 @@
...
+                  reply->payload.body.assign( {'N','U','L','L'});
+
+               context_holder->reply = std::move( reply.value());
+               return 0; // done
+            }
+            else
+            {
+               return -2; // call again
+            }
+         }
```
lazan, line 117: Can we use something other than _magic values_?  `DEFINES` or _enums_?

middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+               casual::http::inbound::call::Context context;
+               http::inbound::call::Request request;
+               http::inbound::call::Reply reply;
+               header::Guard guard;
+            };
+
+            context::Holder* cast( void* holder)
+            {
+               return reinterpret_cast< context::Holder *>( holder);
+            }
```
lazan, line 65: Some `CASUAL_ASSERT` that `holder` points to something?

middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+            { // data storage for single request
+               casual::http::inbound::call::Context context;
+               http::inbound::call::Request request;
+               http::inbound::call::Reply reply;
+               header::Guard guard;
+            };
+
+            context::Holder* cast( void* holder)
+            {
+               return reinterpret_cast< context::Holder *>( holder);
```
lazan: I don't think you need to use `reinterpret_cast` when casting from `void*`. Use `static_cast`

middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+            log::line( verbose::log, "request: ", context_holder->request);
+
+            http::inbound::call::Directive context_directive = directive == 1 ? http::inbound::call::Directive::forward : http::inbound::call::Directive::service;
+
+            context_holder->context = http::inbound::call::Context{ context_directive, std::move( context_holder->request)};
+            // return fd to caller
+            handle->fd = context_holder->context.descriptor().value(); // fd triggered when it's time to poll (when "readable")
+
+            return 0;
```
lazan: If we don't need _return status_, make the function return `void`

middleware/http/source/inbound/c/api.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+               reply->headers.data = context_holder->guard.headers.data;
+               reply->headers.size = context_holder->guard.headers.size;
+
+               auto content_type = common::algorithm::find_if( context_holder->reply.payload.header, []( auto& header)
+               {
+                  return header.key == "content-type";
+               });
+
+               if ( content_type)
```
lazan: `if( content_type)` :)

middleware/http/unittest/source/inbound/test_call.cpp
```diff
@@ -136,8 +139,8 @@ namespace casual
          auto& reply = result.value();
          EXPECT_TRUE( reply.code == http::code::not_found) << CASUAL_NAMED_VALUE( reply.code);
 
-         // TODO: we don't set the content-type on no_entry, is this a problem?
-         EXPECT_TRUE( local::header::value( reply.payload.header, "content-length") == "0");
+         auto size = common::string::compose( "failed to lookup service: ", service, ": TPENOENT").size();
+         EXPECT_TRUE( local::header::value( reply.payload.header, "content-length") == std::to_string( size));
```
lazan, line 143: Interesting check :) It's hard to check the actual payload?  Most likely this check is sufficient though.

middleware/plugin/config
```diff
@@ -1,12 +1,9 @@
...
-platform=`uname`
+. auto/module
 
-if [[ "$platform" -eq "Darwin" ]]; then
-   CORE_LIBS="$CORE_LIBS -L$CASUAL_MAKE_SOURCE_ROOT/middleware/xatmi/bin -lcasual-xatmi -L$CASUAL_MAKE_SOURCE_ROOT/middleware/http/bin -lcasual-http-inbound-common"
-else
-   CORE_LIBS="$CORE_LIBS -L$CASUAL_MAKE_SOURCE_ROOT/middleware/xatmi/bin -Wl,-rpath=$CASUAL_MAKE_SOURCE_ROOT/middleware/xatmi/bin -lcasual-xatmi -L$CASUAL_MAKE_SOURCE_ROOT/middleware/http/bin -Wl,-rpath=$CASUAL_MAKE_SOURCE_ROOT/middleware/http/bin -lcasual-http-inbound-common -Wl,-rpath=$CASUAL_MAKE_SOURCE_ROOT/middleware/common/bin"
-fi
\ No newline at end of file
+ngx_addon_name=$ngx_module_name
```
lazan: Nice, a lot cleaner and easier!

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+
+static void initilize_context( ngx_http_casual_ctx_t* context)
+{
+   context->casual_handle.fd = 0;
+   context->casual_handle.context_holder = NULL;
+}
+
+static ngx_int_t get_header_length( ngx_http_request_t* r)
+{
+   ngx_list_part_t *part = &r->headers_in.headers.part;
```
lazan: `ngx_list_part_t* part`. My brain parser thinks this is easier to read.

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+         {
+            /* The last part, search is done. */
+            break;
+         }
+
+         part = part->next;
+         i = 0;
+      }
+      length++;
+   }
```
lazan: This loop is quite hard to understand. it seems like the outer loop is not really what we want? And why do we loop to increment `length` instead of just add the _elements count_ (`part->nelts`) directly?  Should it not be more like?:
```c
while( part)
{
   length += part->nelts; // might be "of by one"? 
   part = part->next;   
}
```

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+   return length;
+}
+
+static void copy_headers( ngx_http_request_t* r, casual_http_inbound_header_t* header)
+{
+   ngx_list_part_t *part;
+   ngx_table_elt_t *h;
+   ngx_uint_t i;
```
lazan: We should inititalize these... And why not do it closer to usage?

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+    Get the first part of the list. There is usual only one part.
+    */
+   part = &r->headers_in.headers.part;
+   h = part->elts;
+
+   /*
+    Headers list array may consist of more than one part,
+    so loop through all of it
+    */
+   casual_http_inbound_header_t* headerpointer = header;
```
lazan: A assume this is done just to "not effect caller _header_"? It will not be effected, since `header` is passad by value.

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+
+         part = part->next;
+         i = 0;
+      }
+      length++;
+   }
+   return length;
+}
+
+static void copy_headers( ngx_http_request_t* r, casual_http_inbound_header_t* header)
```
lazan: `ngx_http_request_t* request`

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+         part = part->next;
+         h = part->elts;
+         i = 0;
+      }
+
+      headerpointer->key = copy_from_ngx_str_t( h[i].key);
+      headerpointer->value = copy_from_ngx_str_t( h[i].value);
+      headerpointer++;
+   }
+}
```
lazan: Quite hard to understand. More like?: 
```c
ngx_list_part_t* part = &request->headers_in.headers.part;

while( part)
{
   ngx_table_elt_t* element = part->elts;
   
   while( element !=  part->elt + part->nelts)
   {
       header->key =  copy_from_ngx_str_t( element->key);
       header->value =  copy_from_ngx_str_t( element->value);
       ++header;
       ++elements
   }
   part = part->next;
}
```
To me it's not clear how safe it is to increment `header`. How do we now that the memory is allocated and valid?

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+   {
+      ngx_table_elt_t* h;
+
+      /*
+      All we have to do is just to allocate the header...
+      */
+      h = ngx_list_push( &r->headers_out.headers);
+      if (h == NULL) {
+         return NGX_ERROR;
+      }
```
lazan:
```c
ngx_table_elt_t* element  = ngx_list_push( &request->headers_out.headers);
if( ! element) // I think this is valid C
    return NGX_ERROR;
```

middleware/plugin/ngx_http_casual_module.c
```diff
@@ -0,0 +1,688 @@
...
+      */
+      h->hash = 1;
+   }
+    return NGX_OK;
+}
+
+static ngx_int_t buffer_size( ngx_chain_t* response)
+{
+   ngx_int_t result = 0;
+   for( ngx_chain_t *cl = response; cl; cl = cl->next)
```
lazan: `cl`? Is `current` a better name?

lazan (changes requested):
Really good work! A lot less C-stuff, and most of it made sense to me :) 

A few comments I would like you to check out. 

A more general observation:
For me, I tend to "only" use _iterator semantics_ instead of _indexing semantics_.  A lot less _of by one_ bugs, and. easer to reason about (for me at least). And, one should pick one semantics, and not mix them. Since there are "forced" _iterator semantics_ in  `some_list_pointer->next`, the _iterator way_ has been chosen for us :)

middleware/common/source/service/header.cpp
```diff
@@ -16,123 +16,129 @@
 
 namespace casual
 {
-   namespace common
+   namespace common::service::header
```
lazan, line 19: Nice

middleware/http/source/inbound/call.cpp
```diff
@@ -0,0 +1,363 @@
...
+            {
+               auto type( const std::vector< header::Field>& header)
+               {
+                  auto found = algorithm::find( header, "content-type");
+
+                  if( ! found)
+                     common::code::raise::error( code::bad_request, "content-type header is manatory");
+
+                  return protocol::convert::to::buffer( found->value);
+               }
```
lazan, line 53:
```
if( auto found = algorithm::find( header, "content-type"))
    return protocol::convert::to::buffer( found->value);

common::code::raise::error( code::bad_request, "content-type header is manatory");

File: middleware/http/source/inbound/call.cpp [hunk 21]
```diff
@@ -0,0 +1,363 @@
...
+                     default:
+                        return http::code::internal_server_error;
+                  }
+               };
+
+               result.code = transform_http_code( message.code.result);
+               
+               return result;
+            }
+            */
```
lazan: This should be removed?

middleware/http/source/inbound/call.cpp
```diff
@@ -0,0 +1,363 @@
...
+            } // send
+            
+            template< typename Policy>
+            struct basic_caller
+            {
+               basic_caller( communication::ipc::inbound::Device ipc, Request request)
+                  : correlation{ send::lookup( ipc, request.service)}, ipc{ std::move( ipc)}, request{ Policy::transform( std::move( request))}
+               {}
+
+               ~basic_caller()
```
lazan: `~basic_caller() = default;`

Or leave it out, and let the compiler generate one?

lazan (changes requested):
I missed one commit.  I'm not sure how this work, if it's per commit, or the whole PR.

middleware/common/source/service/header.cpp
ahhud: @lazan Just a fun fact, this commit is actually your commit so you´re reviewing yourself which is not a bad idea sometimes either. :-) But I will fix it along with the other commit.

middleware/http/source/inbound/c/api.cpp
ahhud: true

middleware/http/source/inbound/c/api.cpp
ahhud: True, and after all refactoring I think it is used only for one specific type.

middleware/plugin/ngx_http_casual_module.c
ahhud: True, I pretty much stole that example code from tutorials. :-)
I will review all "nginx plugin code" and make it more human readable. :-)

middleware/common/source/service/header.cpp
lazan: I suspected that :)

lazan (dismissed):

lazan (dismissed):

PR #143, opened 2023-02-08
discovery: improve performance by aggregating updates within a duration

lazan:
This patch
* aggregates topology updates within a specific time duration.
* when the duration has passed:
* gather _known_ services/queues (includes "waiting"/_needs_)
* send topology::Explore to all _external_agents_ (outbounds)
* _external_agent_ sends discover::Request to the intersection between the agents known "remote domains" and topology::Explore.domains

This will greatly reduce the amount of messages that is sent. It also improves the "precision" of where discoveries and topology updates is sent, hence only do work that could matter.

### DIFF
--- a/middleware/common/include/casual/platform.h
+++ b/middleware/common/include/casual/platform.h
@@ -164,10 +164,10 @@ namespace casual
                constexpr size::type next = batch::message::pump::next;
             } // message::pump
 
-            namespace topology
+            namespace accumulate
             {
-               constexpr size::type updates = 100;
-            } // topology
+               constexpr auto timeout = std::chrono::seconds{ 1};
+            } // accumulate
             
          } // discovery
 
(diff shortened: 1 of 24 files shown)


middleware/common/include/casual/platform.h
```diff
@@ -168,6 +168,11 @@ namespace casual
             {
                constexpr size::type updates = 100;
```
AlbsAgren: Is this still used now that you've switched to a timeout?

AlbsAgren (dismissed):

middleware/common/include/casual/platform.h
lazan: No, you're right. 

I forgot to reference the issue, I'll remove the constant in the same force push.

AlbsAgren (approved):

PR #151, opened 2023-02-20
Feature/1.6/fix http services listed as discoverable

AlbsAgren:


OlofGullnas:
Noticed that the change of color for #branches seems to be to white. At least judging from naming. I have not checked effect of the escape sequence "\033[0;37m".
Selecting colors can be difficult. What is visible can be difficult to predict. The color setup of the users terminal is not known. Quite common choices are black or white background. White on white is not very visible, nor black on black.

AlbsAgren:
Fair point, a higher contrast color might be a better choice. Before this it was black rather than the grey indicated by the constant, which obivously wasnt easy to read on a terminal with a blackground.

OlofGullnas:
I have used putty for many years. It has:

- default  and bold foreground 
- default and bold background
- and a bunch of ANSI colors (also with bold variants). Black and white exist. 

In Putty the RGB settings for all the colors can be changed in settings...

A similar set of colors seems to be common in terminals that support colors.

I have for many years used some conventions for my terminals: 
- Development systems have either very dark background (almost black, rgb 15, 15, 15 or so) with very light grey default foreground (rgb 240, 240, 240), OR they have the exact opposite Dark text on light background.
- Production system generally have a "pinkish" (low saturation) background  with dark tesxt (red is a "danger" signal...)
- System test (if different from development) and similar have light yellow (low saturation) background with dark foreground (yellow is a "warning" signal)

The idea is to make it obvious when the terminal is connected to a production (or critical in some sense) system. I want to avoid doing bad things by mistake. :-) 

I would avoid using "white" or "black" as colors for text as they can be hard to read with white or black backgrounds. The "deafult" color for text should probably be "default foreground" (is there a standardized escape sequence to select this? It is probably "\033[0m") and never black or white. This is based on the assumption that the users terminal is configured so that default foreground is readable against configured default background.  I also think that "blue" often is hard to read, Especially with a black/dark background , The other colors are better. Yellow can be a little difficult with my yellowish background but usually works with dark and light background.

If white or black is used for text I would recommend setting the background explicitly in the same escape sequence! "033[37;40m" would give white on black background, while "\033[30;47m" would give black on white background. Using "\033[90;107" or \033[97;100m" is also an alternative (bright/bold variants of  white and black).
 
Can  also consider using "reverse" ("\033[7m") in combinaton with colors if more variants are needed! This reverses the role of foreground anf background colors. This should work well with colors that stand out well against both black and white, 

[Wikipedia](https://en.wikipedia.org/wiki/ANSI_escape_code), the section about SGR (Select Graphic Rendition) and some of the following information might be of interest :-)

middleware/administration/unittest/source/cli/test_transaction.cpp
```diff
@@ -111,5 +115,78 @@ namespace casual
          EXPECT_TRUE( output == expected) << "output:  " << output << "expected: " << expected;
       }
 
+      TEST( cli_transaction, list_transactions)
```
lazan, line 118: Nice!

middleware/transaction/source/manager/admin/cli.cpp
```diff
@@ -124,7 +125,15 @@ namespace casual
...
-                        auto format_owner = []( auto& value){ return value.owner.pid;};
+                        auto format_owner = []( auto& value)
+                        {
+                           std::ostringstream out;
+                           if( value.owner.pid)
+                              out << value.owner.pid;
+                           else
+                              out << '-';
+                           return out.str();
+                        };
```
lazan, line 133: You could use 
```cpp
auto format_owner = []( auto& value) -> std::string  
{
   if( value.owner.pid)
     return string::compose( value.owner.pid);
   else
     return "-":      
}
```

lazan (changes requested):
I think the `string::compose` is "better".

middleware/transaction/source/manager/admin/cli.cpp
AlbsAgren: Fixed

lazan (approved):

PR #154, opened 2023-02-23
cli: fix gateway --list-connections to indicate inbound discovery forward

### LINKED ISSUE #147: casual gateway - cli listing of "connections" should show if inbound has discovery forward

lazan:
This patch prints `in*` as bound for inbound with configuration discovery forward = true

resolves #147

### DIFF
--- a/middleware/administration/unittest/source/cli/test_gateway.cpp
+++ b/middleware/administration/unittest/source/cli/test_gateway.cpp
@@ -1,0 +1,104 @@
+//!
+//! Copyright (c) 2022, The casual project
+//!
+//! This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+//!
+
+#include "common/unittest.h"
+
+#include "administration/unittest/cli/command.h"
+
+#include "gateway/unittest/utility.h"
+
+#include "domain/unittest/manager.h"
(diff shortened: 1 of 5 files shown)


middleware/administration/unittest/source/cli/test_gateway.cpp
```diff
@@ -0,0 +1,104 @@
+//!
+//! Copyright (c) 2022, The casual project
```
ahhud, line 2: 2022 -> 2023

ahhud (approved):

middleware/gateway/source/manager/admin/cli.cpp
```diff
@@ -120,6 +120,7 @@ namespace casual
                      {
                         using Enum = decltype( value.bound);
                         case Enum::in: return "in";
+                        case Enum::in_forward: return "in*";
```
AlbsAgren, line 123: How is the meaning of "in*" communicated to the user?

lazan: I was looking for a `--legend` to add this info... There seems to be very little documentation about --list-connections, and I thought that was a separate issue.

PR #159, opened 2023-03-01
unittest - improve stability of several tests

AlbsAgren:
* Fixed a unittest in gateway that would occasionally hang due to a timing error.
* Added a utility function to wait for a service to be advertised. Some tests would fail when fetching a manager's state before that service had been advertised.

### DIFF
--- a/middleware/common/include/common/unittest.h
+++ b/middleware/common/include/common/unittest.h
@@ -75,6 +75,13 @@ namespace casual
          
          // not implemented yet.
          // platform::binary::type receive( const strong::correlation::id& correlation);
+
+         namespace wait::until
+         {
+            //! blocks until `service` has been advertised
+            void advertised( std::string_view service);
+         } // wait::until
+
       } // service
 
 
--- a/middleware/common/source/unittest.cpp
+++ b/middleware/common/source/unittest.cpp
@@ -66,6 +66,24 @@ namespace casual
             return communication::device::blocking::send( lookup.process.ipc, message);
          }
 
+         namespace wait::until
+         {
+            void advertised( std::string_view service)
+            {
+               common::message::service::lookup::Request lookup_request{ process::handle()};
+               lookup_request.requested = service;
+               lookup_request.context.semantic = decltype( lookup_request.context.semantic)::wait;
+               auto correlation = communication::device::blocking::send( communication::instance::outbound::service::manager::device(), lookup_request);
+               communication::ipc::receive< common::message::service::lookup::Reply>( correlation);
+
+               common::message::service::lookup::discard::Request discard_request{ process::handle()};
+               discard_request.correlation = correlation;
+               discard_request.requested = service;
+               discard_request.reply = false;
+               communication::device::blocking::send( communication::instance::outbound::service::manager::device(), discard_request);
+            }
+         } // wait::until
+
       } // service
 
 
--- a/middleware/domain/unittest/source/utility.cpp
+++ b/middleware/domain/unittest/source/utility.cpp
@@ -10,6 +10,8 @@
 
 #include "serviceframework/service/protocol/call.h"
 
+#include "common/unittest.h"
+
 namespace casual
 {
    using namespace common;
@@ -32,6 +34,7 @@ namespace casual
 
       manager::admin::model::State state()
       {
+         common::unittest::service::wait::until::advertised( manager::admin::service::name::state);
          serviceframework::service::protocol::binary::Call call;
          auto reply = call( manager::admin::service::name::state);
          return reply.extract< manager::admin::model::State>();
--- a/middleware/gateway/makefile.cmk
+++ b/middleware/gateway/makefile.cmk
@@ -173,12 +173,6 @@ protocol_objects = [
     make.Compile( 'source/documentation/protocol/example.cpp')
 ]
 
-unittest_archive = make.LinkArchive( 'bin/casual-gateway-unittest',
-    [
-        make.Compile( 'unittest/source/utility.cpp'),
-    ]
-)
-
 # Unittest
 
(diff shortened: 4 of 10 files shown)


lazan (approved):
Looking good.

OlofGullnas:
I built this and tried it: 
On first un with no tests disabled with GTEST_FILTER i got the following errors:
```
[ RUN      ] event_service_log.reopen_log_file
/export/home/guo/casual/middleware/event/unittest/source/service/test_log.cpp:81: Failure
Value of: ! common::unittest::file::fetch::until::content( log_file).empty()
  Actual: false
Expected: true
[  FAILED  ] event_service_log.reopen_log_file (6208 ms)
[ RUN      ] event_service_log.filter_exclusive
[       OK ] event_service_log.filter_exclusive (74 ms)
[ RUN      ] event_service_log.filter_inclusive
/export/home/guo/casual/middleware/event/unittest/source/service/test_log.cpp:158: Failure
Value of: ! common::unittest::file::fetch::until::content( log_file).empty()
  Actual: false
Expected: true
[  FAILED  ] event_service_log.filter_inclusive (6185 ms)
[ RUN      ] event_service_log.filter_inclusive_exclusive
/export/home/guo/casual/middleware/event/unittest/source/service/test_log.cpp:183: Failure
Value of: ! common::unittest::file::fetch::until::content(log_file).empty()
  Actual: false
Expected: true
[  FAILED  ] event_service_log.filter_inclusive_exclusive (6192 ms)
[----------] 4 tests from event_service_log (18660 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (18660 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] event_service_log.reopen_log_file
[  FAILED  ] event_service_log.filter_inclusive
[  FAILED  ] event_service_log.filter_inclusive_exclusive

 3 FAILED TESTS
aborting due to errors

```

I disabled event_service_log with GTEST_FILTER, and tried the unit tests again. Everything runs to completion with no hang (several runs). This is a SIGNIFICANT improvment!  

In the case of event_service_log tests, I  suspect that the added common::unittest::service::wait::until::advertised( manager::admin::service::name::state) is not enough for the unit tests in event_service_log. I guess this is a wait for a service advertised by the domain manager. As I understand it the failures event_service_log are all because entries fail to appear in the event log as expected. I suspect that this is because the event service has not finished its startup when the action that should be logged is executed.

Is there a service advertised by the event service that could be waited for in the event_service_log.* test cases?
I have not looked into how the event service works. Perhaps it needs to "hook inte" the casual machinery to "subscribe" to events. If so waiting for an advertised service might not be enough. Advertising services might happen before subscribing on events...   

I have experimented a bit with thjis before and adding a short sleep (I used 100ms) after the domain startups prevents the errors. The failures are somewhat random. Which tests fail varies between test runs, but it is VERY unusual in my environmnet that all succeed.

lazan:
We could either subscribe to events to wait until the event-logger is spawned. Or, perhaps easier, _fetch domain state until event-logger has running instance_
Either way, it should be an isolated issue.
@OlofGullnas Can you write this one?

OlofGullnas:
I noticed that the Github build on centos failed. Looks like a hang and timeout. Is this the test that PR #160 fixes?

OlofGullnas:
@lazan Would that mean adding something like
 
`auto state = casual::domain::unittest::fetch::until( casual::domain::unittest::fetch::predicate::alias::has::instances( "casual.event-service-log", 1));
`
to the test cases?

lazan:
Exactly, but rather  "casual-event-service-log".

OlofGullnas:
@lazan Did not work. Or rather, the call worked fine, but it did not fix the problem. I suspect this is because hasinstances() returns true as soon as casual-event-service-log has started, but that is still to early.

If I add a small delay it works. A delay of 7 ms gives occasional test failures, 8ms very rarely.  This is with default CASUAL_LOG=(error|warning|information). If I turn on all logging a longer delay is needed. In my system I got occasional failures with a delay of 18ms.
I suspect that the startup of my casual-event-service-log can take a while. My "disk" (home directory) is mounted via NFS and this may slow down things. :-) I tried looking in the log to see if I could get some feeling for timing, but the log gets really big. It did look as if the time from first appearance in the log of casual-event-server-log until I could see messages (that I think are) associated with subscribing to the events was more than 10 ms...

The simple, but very hacky, solution seems to be to go with adding a delay. I can keep the call to "hasinstances()". It does not hurt...

lazan:
> @lazan Did not work. Or rather, the call worked fine, but it did not fix the problem. I suspect this is because hasinstances() returns true as soon as casual-event-service-log has started, but that is still to early.
> 
> If I add a small delay it works. A delay of 7 ms gives occasional test failures, 8ms very rarely. This is with default CASUAL_LOG=(error|warning|information). If I turn on all logging a longer delay is needed. In my system I got occasional failures with a delay of 18ms. I suspect that the startup of my casual-event-service-log can take a while. My "disk" (home directory) is mounted via NFS and this may slow down things. :-) I tried looking in the log to see if I could get some feeling for timing, but the log gets really big. It did look as if the time from first appearance in the log of casual-event-server-log until I could see messages (that I think are) associated with subscribing to the events was more than 10 ms...
> 
> The simple, but very hacky, solution seems to be to go with adding a delay. I can keep the call to "hasinstances()". It does not hurt...

Ok, you've got a _slow_ system :) Or perhaps too fast, in some regards? :) Anyway, the problem is that casua-event-service-log _connects_ to the domain before the log file is created. 
I'll fix this, and then you can remove your sleep.

lazan:
Btw, I don't think that keeping the diskussion in a merged PR is good practice... :) I'll move it to discord...

PR #162, opened 2023-03-04
domain - core servers must restart on unexpected exit

ahhud:
- core servers (singletons) are now set to restart on unexpected exit
- singleton servers logs on error on unexpected exit


lazan (approved):

ahhud:
Merged locally

PR #167, opened 2023-03-11
queue: fix pending data structure slicing

lazan:
Before: When pending objects where moved within the "state machine" some unexpected and unwanted slicing was produced, which lead to "memory leak" (some destructors was not called).

A bit surprising that neither compilers nor clang-tidy catched this bug.

### DIFF
--- a/middleware/queue/include/queue/forward/state.h
+++ b/middleware/queue/include/queue/forward/state.h
@@ -286,7 +286,7 @@ namespace casual
                struct Lookup : transaction_base
                {
                   Lookup() = default;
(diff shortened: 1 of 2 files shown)


AlbsAgren:
We might want to consider adding some performance tests (locust) with longer runtimes to catch future stuff like this?

AlbsAgren (approved):

ahhud (approved):

lazan:
> We might want to consider adding some performance tests (locust) with longer runtimes to catch future stuff like this?

Exactly.

PR #169, opened 2023-03-12
Patch/1.6/some bug fixes

lazan:


ahhud (approved):

OlofGullnas:
Does the failing unit test have a connection to this change? The failing test is 
TEST( common_communication_ipc, block_writes__expect_error_on_send___ok_on_read)

And the failure is in the following sequence:

```
device.connector().block_writes();

         // we can't write any more
         EXPECT_THROW({
            device::non::blocking::send( ipc, unittest::Message{});
         }, std::system_error);

```
I only looked at the Suse build...

OlofGullnas:
After a quick look. My first impression is that blocking writes in this case means changing permissions so that write is no longer allowed for owner, group or other. The first thing I ask myself is if this change of permissions affects already open "files". How does Linux work? It is quite common in other operating systems that permissions are checked on "open", and not on every read and write.

(When I first saw the PR my gut feeling was that this change had to do with "blocking" as opposed to "non-blocking" writes. Obviously something different is meant :-) After checking the commit comment it's obvious that my first thought was wrong... )

OlofGullnas:
Should block_writes() be named disallow_writes() instead? Or perhaps even forbid_writes(). To avoid overuse of "block".

OlofGullnas:
I built this branch on my system with casual-make --debug and casual-make --debug test. The tests executed to completion without errors. An exception is thrown after the ::sendto returns an error.

Why did the test fail in the github build? 

To answer my own question about if a change in permissions affects operations on already open files:  sendto is documented as returning EACCESS for a Unix domain socket when write permission is denied. So the approach used to "block writes" is reasonable.

middleware/common/source/communication/ipc.cpp
```diff
@@ -426,26 +433,42 @@ namespace casual
                      return Handle{ std::move( socket), ipc};
                   }
                } // create 
+
+
+               void block_writes( const std::filesystem::path& path)
+               {
```
OlofGullnas, line 439: I think disallow_writes would be a better name. I associate the words "block" and "blocking" more as the opposite to "nowait".

Other possible name are deny_writes and forbid_writes  Even "prevent_writes" isd a possibility. 

If name is changed, connector::block_writes should also be renamed.

Noticed that Destination::active tests if the owner write bit is set. Suggests  that "deactivate" is also a possible name.

lazan:
We won't do this with domain sockets. We'll fix it when we're using named pipes

PR #172, opened 2023-03-16
configuration - fix queue configuration default directory

### LINKED ISSUE #171: queue configuration
``` 
domain:
  queue:
    default:
      directory: "${CASUAL_DOMAIN_HOME}/queue/groups"
 ``` 
casual does not respect `domain.queue.default.directory`

Either fix it, or remove it from documentation. The latter is somewhat "breaking"

### PR DESCRIPTION
**Pull request by albinsjolund**
Before:
Configuration domain.queue.default.directory was not being used.

Now:
Uses queuebase if any, else <default.directory>/<group.alias>.qb

Fixes #171

### DIFF
--- a/middleware/configuration/source/example/model.cpp
+++ b/middleware/configuration/source/example/model.cpp
@@ -231,7 +231,7 @@ domain:
 
       groups:
          -  alias: A
-            note: "will get default queuebase: ${CASUAL_DOMAIN_HOME}/queue/groupA.gb"
+            note: "will get default queuebase: ${CASUAL_DOMAIN_HOME}/queue/groups/A.qb"
             queues:
                -  name: a1
                -  name: a2
--- a/middleware/configuration/source/model/transform.cpp
+++ b/middleware/configuration/source/model/transform.cpp
@@ -398,13 +398,21 @@ namespace casual
 
                   if( source.groups)
                   {
-                     result.groups = common::algorithm::transform( source.groups.value(), []( auto& group)
+                     auto default_directory = [ &source]() -> std::string
+                     {
+                        if( source.defaults && source.defaults->directory)
+                           return *source.defaults->directory;
+                        return {};
+                     }();
+
+                     result.groups = common::algorithm::transform( source.groups.value(), [ &default_directory]( auto& group)
                      {
                         queue::Group result;
 
                         result.alias = group.alias.value_or( "");
                         result.note = group.note.value_or( "");
                         result.queuebase = group.queuebase.value_or( "");
+                        result.directory = default_directory;
 
                         common::algorithm::transform( group.queues, result.queues, []( auto& queue)
                         {
--- a/middleware/queue/source/group/handle.cpp
+++ b/middleware/queue/source/group/handle.cpp
@@ -586,6 +586,9 @@ namespace casual
                      {
                         if( ! message.model.queuebase.empty())
                            return group::Queuebase{ message.model.queuebase};
+
+                        if( ! message.model.directory.empty())
+                           return group::Queuebase{ message.model.directory + "/" + message.model.alias + ".qb"};
                         
                         auto name = message.model.alias + ".qb";
                         auto file = common::environment::directory::queue() / name;
--- a/middleware/configuration/documentation/build/executable.development.md
+++ b/middleware/configuration/documentation/build/executable.development.md
@@ -4,13 +4,13 @@
 
 Defines user configuration when building a casual executable.
 
-## resources
+### resources
 
 Defines which `xa` resources to link and use runtime. A name **has** to be provided for each 
 resource, startup configuration phase will ask for resource configuration for that 
 given name.
 
-## entrypoint
+### entrypoint
 
 Defines the name of the user provided _entry point_. The signature has to be the same as a
 normal main function `int <entrypoint-name>( int argc, char** argv)`.
--- a/middleware/configuration/documentation/build/server.development.md
+++ b/middleware/configuration/documentation/build/server.development.md
@@ -4,7 +4,7 @@
 
 Defines user configuration when building a casual server.
 
-## services
+### services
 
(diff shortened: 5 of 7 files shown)

### CODE REVIEW

**Comment by OlofGullnas** [comment]
I tried this branch.  Wanted to look at a few things. Built OK, but casual-make --debug test failed! Surprised me a bit as the tests went OK in the Github build.

Failed in the test casual_queue_forward.configuration_post.

Built patch/1.6/main instead (parent to this PR). It also built OK, and test suite completed without errors.
Went back to patch/1.6/ queue_configuration branch. Same test fails. When looking at the test printout it seems as if the changes applied in the "post" done in the test are missing (i.e. was not applied). The x1, x2 and x3 queues did not appear in the printout of udated.queue.

I checked casual.log and the "suspicious" thing I notice are the entries:

```
2023-03-17T12:30:22.482769+01:00|forward-domain|201deb530ccd436b95f6faee878d035d|109333|139662476830592|casual-domain-manager||||error|process cored, alias: extra, details: { pid: 109348, status: 0, reason: core}
2023-03-17T12:30:22.486170+01:00|forward-domain|201deb530ccd436b95f6faee878d035d|109333|139662476830592|casual-domain-manager||||information|process exited, alias: extra, details: { pid: 109348, status: 0, reason: core}
```
The executable that cores seems to be middleware/queue/bin/casual-queue-group. At least "sudo coredumpctl" lists a core produced by that executable...

As the "previous" commit does not have this problem, it is likely that the changes in this PR are involved. 

I tried gdb on the core file. It fails to produce a backtrace (goes into a "loop", possibly because of missing debug information) but the top of the stack displays as:

```
(gdb) bt
#0  0x00007fb49340a6d9 in ?? () from /usr/lib64/libsqlite3.so.0
#1  0x00007fb49340a865 in ?? () from /usr/lib64/libsqlite3.so.0
#2  0x00007fb49340a8f8 in ?? () from /usr/lib64/libsqlite3.so.0
#3  0x00007fb493446ad8 in ?? () from /usr/lib64/libsqlite3.so.0
#4  0x00007fb4934bd01f in ?? () from /usr/lib64/libsqlite3.so.0
#5  0x00000000004ac672 in sql::database::Connection::open (file=filesystem::path "/extra.qb" = {...})
    at ../../thirdparty/database/include/sql/database.h:649
#6  0x00000000004ac7e7 in sql::database::Connection::open (file=filesystem::path "/extra.qb" = {...})
    at ../../thirdparty/database/include/sql/database.h:664
#7  0x00000000004ac7e7 in sql::database::Connection::open (file=filesystem::path "/extra.qb" = {...})
    at ../../thirdparty/database/include/sql/database.h:664

```

A path of just "/extra.qb" is a bit suspicious :-)

**Comment by OlofGullnas** [comment]
Much better after last commit. No longer fails the test suite. Will try to take a look at how things now work. That is if the queuebase filename handling is reasonable :-). May take a day or two (or three). Busy with building work over the weekend.

File: middleware/configuration/source/model/transform.cpp [hunk 1]
```diff
@@ -398,13 +398,20 @@ namespace casual
...
+                     }();
+
+                     result.groups = common::algorithm::transform( source.groups.value(), [ &defaults]( auto& group)
                      {
                         queue::Group result;
 
                         result.alias = group.alias.value_or( "");
                         result.note = group.note.value_or( "");
-                        result.queuebase = group.queuebase.value_or( "");
+                        result.queuebase = group.queuebase.value_or( defaults != "" ? defaults + "/" + result.alias + ".qb" : "");
```
lazan: I'm not sure that _configuration_ should be responsible for the queuebase name. At least the decision should be made in one place.
I think the _queue module_ should have the responsibility. If so, we need to extend the model.queue.group with _directory_, and make the decision when configure the queue-group with the model.

middleware/configuration/source/example/model.cpp
```diff
@@ -231,7 +231,7 @@ namespace casual
 
       groups:
          -  alias: A
-            note: "will get default queuebase: ${CASUAL_DOMAIN_HOME}/queue/groupA.gb"
+            note: "will get default queuebase: ${CASUAL_DOMAIN_HOME}/queue/groups/A.qb"
```
lazan, line 234: Nice! You need to regenerate the documentation though. There's a _generator_ under bin that you should use. `--help` should be sufficient to understand.

OlofGullnas:
While looking around at queue configuration related things I noticed that  domain.queue.operation.md gives the wrong default location for queuebase. It gives `${CASUAL_DOMAIN_HOM}/queue/<group-name>.qb`
as the default (the env var is incorrectly spelled...). The actual default seems to be `${CASUAL_QUEUE_DIRECTORY}/<group-name>.qb`, CASUAL_QUEUE_DIRECTORY in turn defaults to `${CASUAL_PERSISTENT_DIRECTORY}/queue/` And CASUAL_PERSISTENT_DIRECTORY defaults to `$(CASUAL_HOME_DIRECTORY}/.casual`. If only "mandatory" environment variables are defined the end result is `${CASUAL_DOMAIN_HOME}/.casual/queue/<group-name>.qb`.

I think the documented default is the "old" default location! Before is was changed to be under ".casual". There is some code in configuration::update::detail::queuebase() to move the ".qb" file from that location to the new default location. This code is only used if neither "quebase:" nor "directory:" are specified (i.e. the location of queue database files are completely defaulted). 

Should perhaps be a separate issue, but as there already are updates to queue related things in this .md file in this PR...

middleware/queue/source/group/handle.cpp
```diff
@@ -586,6 +586,9 @@ namespace casual
                      {
                         if( ! message.model.queuebase.empty())
                            return group::Queuebase{ message.model.queuebase};
+
+                        if( ! message.model.directory.empty())
+                           return group::Queuebase{ message.model.directory + "/" + message.model.alias + ".qb"};
```
lazan, line 592: Nice!

lazan (approved):

PR #174, opened 2023-03-24
queue - send error reply on failed enqueue

### LINKED ISSUE #173: no error reply on failed enqueue
When an enqueue fails, e.g. when the database is full, no error reply will be sent, causing whomever requested the enqueue to wait forever.

AlbsAgren:
Send an error reply was sent when enqueues fail, such as when the queuebase runs out of memory

Fixes #173

### DIFF
--- a/middleware/queue/source/group/handle.cpp
+++ b/middleware/queue/source/group/handle.cpp
@@ -229,6 +229,7 @@ namespace casual
                         catch( ...)
                         {
                            log::line( log::category::error, exception::capture(), " failed with enqueue request to queue: ", message.name);
+                           state.multiplex.send( message.process.ipc, common::message::reverse::type( message));
                         }
                      };
                   }
--- a/middleware/queue/unittest/source/test_queue.cpp
+++ b/middleware/queue/unittest/source/test_queue.cpp
@@ -6,6 +6,7 @@
 
 
 #include "common/unittest.h"
+#include "common/unittest/file.h"
 
 #include "queue/unittest/utility.h"
 #include "queue/common/queue.h"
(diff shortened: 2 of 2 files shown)


middleware/queue/source/group/handle.cpp
```diff
@@ -229,6 +229,7 @@ namespace casual
                         catch( ...)
                         {
                            log::line( log::category::error, exception::capture(), " failed with enqueue request to queue: ", message.name);
+                           state.multiplex.send( message.process.ipc, common::message::reverse::type( message));
```
OlofGullnas, line 232: I had expected some kind of error code to be set to indicate what happened. Appears as if there is no way to return an error in the reply, except by not providing a "quid" in message.queue of the reply.

OlofGullnas (approved):
No indication of why the enqueue failed is returned in the reply message. Adding this is probably not needed in the short term, and is a bigger change. The only indication of what really happened seems to be in the logs.

This change resolves issue #173 and prevents the "wait forever" reported there so it's "good enough".

middleware/queue/unittest/source/test_queue.cpp
```diff
@@ -1222,5 +1224,41 @@ namespace casual
...
+
+         queue::Message message;
+         message.payload.type = common::buffer::type::binary;
+         message.payload.data = common::unittest::random::binary( 512);
+
+         EXPECT_THROW( common::algorithm::for_n( 2000, [ &message]()
+         {
+            queue::enqueue( "a1", message);
+         }), std::system_error);
+      }
```
lazan, line 1261: Good test case

lazan (approved):
Nice!

PR #176, opened 2023-04-07
Patch/1.6/some bugs

lazan:


middleware/common/source/transcode.cpp
ahhud, line 1: Could this be a prospect to some separate core functionality in the future? 
It offer core functionality, but has nothing to do with casual. Just like the `argument` functionality.

middleware/common/include/common/transcode.h
```diff
@@ -17,266 +17,296 @@
...
+         namespace detail
+         {
+            template< typename Input, typename Out>
+            void encode( Input first, Input last, Out out)
             {
-               auto cast = []( auto&& i){ return reinterpret_cast< const char*>( &(*i));}; 
-               detail::encode( out, cast( first), cast( last));
-               return out;
-            }
+               auto hex = []( auto value)
```
ahhud, line 175: Are the meaning of these "magic" numbers obvious when calculating hex?

middleware/common/include/common/transcode.h
```diff
@@ -17,266 +17,296 @@
...
+                     return value + 48;
+                  return value + 87;
+               };
 
-            template< typename R>
-            auto encode( std::ostream& out, R&& range) -> 
-               std::enable_if_t< traits::is::binary::like_v< R>, std::ostream&>
-            {
-               return encode( out, std::begin( range), std::end( range));
+               for( ; first != last; ++first)
```
ahhud, line 182: Is this some "standard" formula when encoding? Looks a little like voodoo to me. ;-)

ahhud (approved):
Just some curious remarks, not affecting the review. :-)

middleware/common/include/common/transcode.h
lazan: I would not say it's  _obvious_ :) But it's not that difficult, take the upper 4 bits and make a character (we need to adjust to make it a valid hex char, 10 = a, and so on), then take the lower 4 bits.  That is, one byte -> two _chars_ with values [0-f].  
We might be able to use some  c++ _format_ stuff instead of hand rolling it... But I'm yet to find a suitable replacement.

middleware/common/include/common/transcode.h
lazan: It's not that voodoish :)  bitmask the upper 4 bits -> right shift 4 places (bits) (so it becomes the lower 4 bits, hence has a value of 0 to 2 ^ 4) -> transform the value to a hex-friendly char.  Do the same for the lower 4 bits, but we don't need to bit-shirt any thing, since the 4 bits are already in the "right" place.

OlofGullnas: Works as long as the character sets are "ASCII" based. Would not work with EBCDIC. In practice only used in IBM mainframes (as far as I know. May be some other odd systems out there) so I do not think we need to worry too much :-)
EBCDIC has digits starting at 0xF0, and A-Z (and corresponding lower case characters) are split into 3 separate ranges, and lower case 'a' is not 0x61.

middleware/common/source/transcode.cpp
lazan: Maybe, but one can argue that a lot of stuff in common offers _core functionality_. For me, it's more important to try to separate and move stuff out from common that are "more casual specific". The "xatmi server" stuff, the call stuff, the transaction stuff. 
Naively: 
transaction stuff -> transaction (no brainer :))
"xatmi server" stuff -> som new "server" component.
"call stuff" -> maybe to the same "server", or it's own. (don't need to be a "server" to be able to call stuff)

I've given it a try a few times, but it didn't feel right with the things I tried.

PR #183, opened 2023-04-25
queue - prevent forward causing service-manager coredump

### LINKED ISSUE #182: casual-queue-forward sometimes causes the service manager to core

AlbsAgren:
Before: Occasionally, the forward would request a discard of a service lookup but still call the service, depending on the order of the replies. This would trigger a core in the service-manager when the service ACK:ed the call.

Now: We handle pending service discard requests separately so that no service call occurs irrespective of the order of replies.

Fixes #182

### DIFF
--- a/middleware/queue/source/forward/handle.cpp
+++ b/middleware/queue/source/forward/handle.cpp
@@ -320,12 +320,14 @@ namespace casual
                         
                         // we assume service-manager are always 'online'...
                         ipc::flush::send( ipc::service::manager(), request);
+
+                        state.pending.service.lookup_discards.emplace_back( std::move( pending));
                      };
 
-                     auto& pending = state.pending.service.lookups;
+                     // we consume all pending service lookups
+                     auto pending = std::exchange( state.pending.service.lookups, {});
                      algorithm::for_each( pending, send_service_lookup_discard_request);
                   }
-
 
                   // start the "flows" for configured forwards
                   auto lookup_request = []( auto& state, auto& forwards)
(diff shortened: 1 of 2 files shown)


middleware/queue/source/forward/handle.cpp
```diff
@@ -320,13 +320,15 @@ namespace casual
                         
                         // we assume service-manager are always 'online'...
                         ipc::flush::send( ipc::service::manager(), request);
+
+                        state.pending.service.lookup_discards.emplace_back( std::move( pending));
                      };
 
-                     auto& pending = state.pending.service.lookups;
+                     // we consume all pending service lookups
+                     auto pending = std::exchange( state.pending.service.lookups, {});
```
lazan, line 328: Beautiful.

middleware/queue/source/forward/handle.cpp
```diff
@@ -575,8 +577,15 @@ namespace casual
                            if( message.state == decltype( message.state)::busy)
                               return;
 
+                           // if we cant find a pending lookup we assume the lookup is discarded and handle it later
+                           auto found = common::algorithm::find( state.pending.service.lookups, message.correlation);
+                           if( ! found)
+                           {
+                              log::line( verbose::log, "no pending lookup found for correlation: ", message.correlation);
+                              return;
+                           }
```
lazan, line 588: When we reach this, there should be a pending _lookup::discard_, should we check this and log error if not? 
```
if( ! algorithm::find( state.pending.service.discard_lookup, message.correlation)
    log::line( log::category::error, code::casual::invalid_semantics, " expected descard_lookup state for correlation: ", message.correlation);
```
It sure would help us, if for some reason this state is reached. 
Or, maybe we can rely on the discard reply any way... Hmm, for me it feels like a good thing that every _handler_ sort of taker care of the expected state. In this case, if it's not in pending lookups, it should be in pending discard, if so, I'm good. But my thinking might be overkill..

lazan (approved):
Contemplate my comment about verifying expected state. I think it's good as is though.

middleware/queue/source/forward/handle.cpp
AlbsAgren: I guess we might as well throw it in for good measure, though since the error is mainly aimed at 'us' (as in the developers) maybe it should be an assert in order to be symmetrical with pending::consume?

lazan: > I guess we might as well throw it in for good measure, though since the error is mainly aimed at 'us' (as in the developers) maybe it should be an assert in order to be symmetrical with pending::consume?

Good point. Yeah, i think it's hard to change the consume to an _error log_ (which might be better in retrospect), an assert it is then.

lazan:
Merged manually

Issue #188, opened 2023-05-12
gateway - always advertise discovered resources


ahhud:

ahhud:
Right now outbound gateway only advertise changes to its list of known resources, causing a non robust semantic if it´s customers e.g. casual-queue-manager has restarted.

PR #190, opened 2023-05-12
gateway - always advertise discovered resources (#188)

ahhud:
Before: On a disovery, outbound only advertise new services and queues resulting in problems e.g. if casual-queue-manager has lost it´s state from a restart.

After: Outbound always advertises and let 'the customers' know about found resources.

### DIFF
--- a/middleware/gateway/unittest/source/test_manager.cpp
+++ b/middleware/gateway/unittest/source/test_manager.cpp
@@ -548,6 +548,96 @@ domain:
          }
       }
 
+      TEST( gateway_manager,  enqueue_dequeue___restart_queuemanager_expect_message)
+      {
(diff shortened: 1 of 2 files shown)


middleware/gateway/unittest/source/test_manager.cpp
```diff
@@ -548,6 +548,96 @@ namespace casual
...
+                  {
+                     return done;
+                  })
+               ),
+               [&done]( const common::message::event::process::Spawn& event)
+               {
+                  log::line( log::debug, "event: ", event);
+                  done = event.alias == "casual-queue-manager";
+               });
+         }
```
lazan, line 617: Nice.

lazan (approved):

PR #193, opened 2023-05-16
service - always send timeout duration in lookup reply

AlbsAgren:
The service manager now always includes the configured timeout duration in lookup replies even if the service is not timoutable.

### DIFF
--- a/middleware/common/include/common/message/service.h
+++ b/middleware/common/include/common/message/service.h
@@ -115,7 +115,7 @@ namespace casual
                inline auto logical_name() const noexcept { return requested.value_or( name); }
 
                CASUAL_CONST_CORRECT_SERIALIZE(
-                  service::Base::serialize( archive);
+                  message::Service::serialize( archive);
                   CASUAL_SERIALIZE( requested);
                )
             };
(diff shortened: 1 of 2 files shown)


middleware/common/include/common/message/service.h
```diff
@@ -115,7 +115,7 @@ namespace casual
                inline auto logical_name() const noexcept { return requested.value_or( name); }
 
                CASUAL_CONST_CORRECT_SERIALIZE(
-                  service::Base::serialize( archive);
```
AlbsAgren, line 118: This took more than a second to find 😅

aliothe:
Betyder det här även att man skickar med timeout:en i service callet?

AlbsAgren:
Ja, anroparen får timeoutvärdet från lookupen.

aliothe:
https://github.com/casualcore/casual-java/blob/dev/casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java#L124

Dvs casual jca inbound använder sig av timeout:en som kommer i service callet.

aliothe:
Så så länge det kommer där så kommer det att användas.

middleware/common/include/common/message/service.h
lazan: Ahh, subtle! Good catch!

lazan (approved):
Nice.
No unittest? Is it not worth it?

AlbsAgren:
I thought about a unittest but it felt like it would end up being more complex than I thought was called for, but I could be convinced otherwise.

PR #195, opened 2023-05-17
Patch/1.6/multiplex send recreate

lazan:


middleware/service/source/manager/state.cpp
```diff
@@ -66,11 +66,15 @@ namespace casual
 
             void Sequential::unreserve( const common::message::event::service::Metric& metric)
             {
-               assert( state() == State::busy);
```
AlbsAgren, line 69: Good riddance, i think

AlbsAgren (approved):

Issue #197, opened 2023-05-22
outbound gateway contains services with incorrect routing


ahhud:
Sometimes the internal state of outbound gateway contains routing (lookup) from services to descriptors with values not pointing to any live connection (perhaps old disconnected connections).
Sometimes the descriptors also pointing to live but completely wrong connections.

One thesis is that there is a flaw in cleanup in old connections causing stale references to old descriptors/connections and when new connections appear and reuse the old descriptor value it causes a false coupling between a service and a connection which leads to misrouting av calls.

In QA I have seen that the outbound state also contained references to 4 pending transactions coupled to a stale descriptor causing a failure with the message "failed to correlate descriptor" when trying to send a message.

PR #203, opened 2023-05-25
gateway - handle routing problems (#197)

ahhud:
Before:
When discovery requests still beeing in-flight and only some of the destinations had replied and a connection lost was received on a connection that already had replied, that reply still were wrongly considerd as a valid provider of the requested service. That caused faulty entries in the service lookup list.

After:
When a connection lost is received all replies in an ongoing discovery from that connection/descriptor are considered failed.

### DIFF
--- a/middleware/common/include/common/message/coordinate.h
+++ b/middleware/common/include/common/message/coordinate.h
@@ -93,24 +93,23 @@ namespace casual
                   return done();
                }
 
-               template< typename I>
-               auto failed( I&& id) -> std::vector< strong::correlation::id>
+               template< typename I, typename Predicate>
+               auto failed( I&& id, Predicate predicate) -> std::vector< strong::correlation::id>
                {
-                  auto has_id_and_pending = [ &id]( auto& pending){ return pending.id == id && pending.state == Pending::State::pending;};
-                  auto update_failed_and_transform_correlation =  []( auto& pending)
-                  {
-                     pending.state = Pending::State::failed; 
-                     return pending.correlation;
-                  };
-
-                  return algorithm::transform_if( m_pending, update_failed_and_transform_correlation, has_id_and_pending);
+                  return algorithm::transform_if( m_pending, []( auto& pending)
+                     {
+                        pending.state = Pending::State::failed;
+                        return pending.correlation;
+                     },
+                     predicate( id)
+                  );
                }
 
                bool done()
                {
                   if( algorithm::any_of( m_pending, predicate::value::equal( Pending::State::pending)))
                      return false;
-                  
+
                   m_callback( std::move( m_received), std::move( m_pending));
                   return true;
                }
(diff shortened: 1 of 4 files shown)


middleware/gateway/unittest/source/test_manager.cpp
```diff
@@ -618,7 +618,7 @@ namespace casual
 
          // because of some unknown factor, the queue-manager needs some time to get ready
          // to handle calls. Strange but is a separate issue.
-         process::sleep( std::chrono::milliseconds{ 10});
+         process::sleep( std::chrono::milliseconds{ 500});
```
AlbsAgren, line 621: Is this similar to the issue where some unittests had services being called before they were advertised that we dealt with a few months back?

middleware/common/include/common/message/coordinate.h
```diff
@@ -93,19 +93,26 @@ namespace casual
...
-                  {
-                     pending.state = Pending::State::failed; 
-                     return pending.correlation;
-                  };
-
                   return algorithm::transform_if( m_pending, update_failed_and_transform_correlation, has_id_and_pending);
                }
 
+               template< typename I>
+               auto purge( I&& id) -> std::vector< strong::correlation::id>
```
OlofGullnas: Is "purge" a good name for this function? As I understand it this function does not purge anything...
I think of purge as a verb with meaning "get rid of" or "discard". The function returns correlation ids for items _**to be**_ purged.  Would to_purge be a better name? The purge function below in the code does purge items (with the help of lookup_erase())

middleware/gateway/unittest/source/test_manager.cpp
lazan: Yes, sort of the same thing.

middleware/common/include/common/message/coordinate.h
```diff
@@ -93,19 +93,26 @@ namespace casual
...
                auto failed( I&& id) -> std::vector< strong::correlation::id>
                {
                   auto has_id_and_pending = [ &id]( auto& pending){ return pending.id == id && pending.state == Pending::State::pending;};
-                  auto update_failed_and_transform_correlation =  []( auto& pending)
-                  {
-                     pending.state = Pending::State::failed; 
-                     return pending.correlation;
-                  };
-
                   return algorithm::transform_if( m_pending, update_failed_and_transform_correlation, has_id_and_pending);
```
lazan: `update_failed_and_transform_correlation` is a regular function. Use  `&update_failed_and_transform_correlation` to make it more clear that we're taking the adress to the function. Hence, not relying on implicit conversion to function pointer.
Not super important, but I think it improves local reasoning to easer see that we're using a function instead of some other callable.

middleware/common/include/common/message/coordinate.h
lazan: I concur. I think @ahhud realised that we can't really remove stuff "yet" here. We need to call the _done_ to possible trigger user callback later. 
Another name would be preferable. Or combine the two and provide the predicate for which the "filtering" is done.

lazan (changes requested):

ahhud:
@lazan @AlbsAgren @OlofGullnas I made some requested changes and force-pushed. 
Hm, I think something became strange with that force-push

ahhud:
@lazan @AlbsAgren @OlofGullnas Starting to go cross-eyed, but now is the correct reworked version force-pushed.

middleware/common/include/common/message/coordinate.h
```diff
@@ -93,24 +93,23 @@ namespace casual
...
-
-                  return algorithm::transform_if( m_pending, update_failed_and_transform_correlation, has_id_and_pending);
+                  return algorithm::transform_if( m_pending, []( auto& pending)
+                     {
+                        pending.state = Pending::State::failed;
+                        return pending.correlation;
+                     },
+                     predicate
+                  );
                }
```
lazan, line 106: `id` is not used. Either remove it from the arguments, or invoke the predicate with it. If you let the predicate take the `id` as a parameter you don't need to capture it in the lambda (creation of the predicate) further down. I prefer the latter, but the former is good too.

lazan (changes requested):
Nice work.

lazan (approved):

Issue #198, opened 2023-05-22
outbound gateway only discover some of the the domains containing an service


ahhud:
If domain A is connected like a hub to 20 other domains each containing the same 5 services (S1, S2, S3, S4, S5), domain A then discover S1 in 20 domains, S2 in 18, S3 in 20, S4 in 19 and S5 in 17 domains, i.e. somewhat random.
The expectation is to find all 5 services in all 20 connected domains.

If one dosen't do that it can cause transaction problems.

PR #206, opened 2023-05-26
discovery - handle prospect services (#198)

ahhud:
Before:
In-flight discovery request was not considered upon rediscovery requests and explore resulting in missing services ( and queues) routing to new connections.

After:
Every in-flight discovery request registers it's resources to a prospect list which is considered when doing rediscover and exlpore. After the discovery request is done all prospect claims are removed from the prospect list made by that correlation.

### DIFF
--- a/middleware/domain/source/discovery/state.cpp
+++ b/middleware/domain/source/discovery/state.cpp
@@ -139,6 +139,40 @@ namespace casual
          return false;
       }
 
-      
+      void State::Prospects::insert( const Content& content, common::strong::correlation::id correlation)
+      {
+         for( auto service : content.services)
+         {
+            services[ service].emplace_back( correlation);
+         }
+         for( auto queue : content.queues)
+         {
+            queues[ queue].emplace_back( correlation);
+         }
+      }
+
+      void State::Prospects::remove( common::strong::correlation::id correlation)
+      {
+         algorithm::container::erase_if( services, [ &correlation]( auto& pair) 
+         {
+            return algorithm::container::erase( pair.second, correlation).empty();
+         });
+
+         algorithm::container::erase_if( queues, [ &correlation]( auto& pair) 
+         {
+            return algorithm::container::erase( pair.second, correlation).empty();
+         });
+      }
+
+      State::Prospects::Content State::Prospects::content() const
+      {
+         State::Prospects::Content content;
+         algorithm::transform( services, content.services, []( auto& pair) { return pair.first;});
+         algorithm::transform( queues, content.queues, []( auto& pair) { return pair.first;});
+         algorithm::container::sort::unique( content.services);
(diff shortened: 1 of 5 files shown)


middleware/domain/source/discovery/state.cpp
```diff
@@ -139,6 +139,42 @@ namespace casual
...
+         }
+      }
+
+      void State::Prospects::remove( common::strong::correlation::id correlation)
+      {
+         for( auto& service : services)
+         {
+            algorithm::container::erase( service.second, correlation);
+         }
+         algorithm::container::erase_if( services, []( auto pair) { return pair.second.empty();});
```
lazan: You could erase the correlation in the same algorithm, hence only do one pass.

```suggestion
         algorithm::container::erase_if( services, [ &correlation]( auto& pair) 
         {
             return algorithm::container::erase( pair.second, correlation).empty();
         );
```
And you need to take the pair by reference.

middleware/domain/source/discovery/state.cpp
```diff
@@ -139,6 +139,42 @@ namespace casual
...
+            algorithm::container::erase( queue.second, correlation);
+         }
+         algorithm::container::erase_if( queues, []( auto pair) { return pair.second.empty();});
+      }
+
+      State::Prospects::Content State::Prospects::content() const
+      {
+         State::Prospects::Content content;
+         algorithm::transform( services, content.services, []( auto pair) { return pair.first;});
+         algorithm::transform( queues, content.queues, []( auto pair) { return pair.first;});
```
lazan: Take `pair` by reference.

lazan (changes requested):
A general note: Are you not worried about logging state in a few places? Since we got the option of dumping the state by `internal --state-dump <pid>`. Maybe _discovery_ does not generally contain that much state?  It gets "harder" to read the loggs in my opinion, on the other hand, all state is logged which could be very helpful :) I dunno...

ahhud:
> A general note: Are you not worried about logging state in a few places? Since we got the option of dumping the state by `internal --state-dump <pid>`. Maybe _discovery_ does not generally contain that much state? It gets "harder" to read the loggs in my opinion, on the other hand, all state is logged which could be very helpful :) I dunno...

It was a very small state. But you are correct, I will remove it.

middleware/domain/source/discovery/state.cpp
ahhud: @lazan So it is safe to remove elements from _services_ when inside a range for-loop over _services_?

ahhud: Now I got it.

ahhud:
@lazan I made the proposed changes and force-pushed

lazan (approved):

PR #208, opened 2023-05-31
domain - handle shutdown signals

AlbsAgren:
Whether the unittest actually provides any value is debatable, see what you think.

### DIFF
--- a/middleware/common/unittest/source/message/test_dispatch.cpp
+++ b/middleware/common/unittest/source/message/test_dispatch.cpp
@@ -8,19 +8,23 @@
 
 #include <common/unittest.h>
 
-
-
+#include "common/signal.h"
+#include "common/process.h"
 
 #include "common/message/service.h"
 #include "common/message/server.h"
 #include "common/message/dispatch.h"
 #include "common/message/transaction.h"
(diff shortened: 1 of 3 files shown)


middleware/common/unittest/source/message/test_dispatch.cpp
```diff
@@ -73,6 +77,23 @@ namespace casual
...
+            throw exception::compose( code::casual::interrupted);
+         }
+         catch( ...)
+         {
+            std::optional< std::system_error> error;
+            EXPECT_NO_THROW( error = message::dispatch::condition::detail::handle::error());
+
+            EXPECT_TRUE( error && error->code() == code::signal::terminate);
+         }
+      }
```
lazan, line 95: This tell us that we don't throw, and we get the signal as an error, as expected. I think this is good enough.

lazan (changes requested):
I think you want to handle the  `code::signal::terminate` in the _error condition_ in domain manager, as a "nice path". Otherwise we'll log error on sigterm, which we don't want?

Otherwise, good work.

lazan (approved):

PR #210, opened 2023-06-01
discovery: aggregate external discovery requests for performance

lazan:
This patch tries to aggregate discovery::Request based on load. The load is measured as _current discovery request in-flight_.

We've got a _load-window_ that defines how many in-flights to trigger the "next level". Every "level" adds a _timeout_ where every external discovery request are accumulated within.

When the _timeout_ kicks in, the aggregated request is sent, and later when the accumulated reply arrives it is _fan out_ to corresponding replies to match the actual original requests (correlation id and such).

"level" semantics:

The first "level" is `0`, hence has no accumulation and no timeout -> every request is sent "as is".

"level" 1 has a timeout with accumulation until send of 1 x _timeout value_.

"level" 2 -> 2 x _timeout value_.

... and so on.

To reach the "next level" the load has to increase with the _load-window_.

Example: Lets say _load-window_ is 2, and _timeout value_ is 10ms. If we have a load of 3, a timeout is set to 10ms, and we start accumulating request. When the timeout kicks in, we send the request, and invalidate the timeout.
A new request is about to be "sent", we check the load and decide if we should accumulate or not. We could have got replies during this time, and the load has gone down.

Note: To reach "level 2", there need to be a load of 5 -> including at least 2 accumulated requests -> at least 2 x 2ms without "any" replies.

This "algorithm" should self tune against heavy loads, at lest to other domains over network.

Temporary "config", will probably be renamed or removed in the future:
* CASUAL_DISCOVERY_ACCUMULATE_REQUESTS = <load window default 20>
* CASUAL_DISCOVERY_ACCUMULATE_TIMEOUT = <timeout default 4ms>

### DIFF
--- a/middleware/common/include/casual/platform.h
+++ b/middleware/common/include/casual/platform.h
@@ -166,7 +166,8 @@ namespace casual
 
             namespace accumulate
             {
-               constexpr auto timeout = std::chrono::seconds{ 1};
+               constexpr size::type requests = 20;
+               constexpr auto timeout = std::chrono::milliseconds{ 4};
             } // accumulate
             
          } // discovery
--- a/middleware/domain/include/domain/discovery/state.h
+++ b/middleware/domain/include/domain/discovery/state.h
@@ -173,7 +173,7 @@ namespace casual
 
          namespace accumulate
          {
-            namespace topology::state
+            namespace topology
             {
                struct Direct
                {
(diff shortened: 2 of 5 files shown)


ahhud (approved):
I think there is some unnecessary state.prospect.remove as a result of the changes and some incorrect trace texts in handle.cpp. But it is very minor so I approve and releases.

PR #211, opened 2023-06-02
service - include parent in service metrics

### LINKED ISSUE #209: service metrics does not always include parent

AlbsAgren:
This commit fixes an issue where the parent of a service call would get lost in the resulting metric.

Resolves #209


lazan (approved):
Nice catch1

PR #213, opened 2023-06-03
discovery: replies should only contain a subset of the requested

lazan:
Before: when we fan out the replies for accumulated discovery replies from other domains we do not filter base on the requested services and queues. Hence, the replies to caller could contain services and queues that was not asked for. Which lead to domains knows about services/queues that there hasn't been a need for.

Now: The accumulated reply content is intersected with what the actual caller has requested -> a subset of the requested is replied to caller.

### DIFF
--- a/middleware/domain/include/domain/discovery/state.h
+++ b/middleware/domain/include/domain/discovery/state.h
@@ -393,17 +393,19 @@ namespace casual
                   //! adds the content and associate with the correlation.
                   void add( const common::strong::correlation::id& correlation, const request_content& content);
 
+                  //! adds partial "local" known 'content', that will be used in `filter_reply`
+                  void add_known( const common::strong::correlation::id& correlation, reply_content&& content);
+
                   template< typename M>
                   void add( const M& message) { add( message.correlation, message.content);}
 
                   //! adds all cached resources to services and queues
                   request_content complement( request_content&& content);
 
-                  //! @returns the "cached" requested resources associated withe the correlation intersected with the 
-                  //! supplied content. Also cleans the associated state. 
-                  //! @attention this function is not idempotent.
+                  //! @returns the "cached" requested resources associated with the correlation intersected with the 
+                  //! supplied content. Also adds partial known "local" content, if any. 
+                  //! @attention Cleans the associated state -> this function is not idempotent.
                   reply_content filter_reply( const common::strong::correlation::id& correlation, const reply_content& content);
-
 
                   CASUAL_LOG_SERIALIZE(
                      CASUAL_SERIALIZE( m_services);
@@ -413,6 +415,7 @@ namespace casual
                private:
                   cache::Mapping m_services;
                   cache::Mapping m_queues;
+                  std::unordered_map< common::strong::correlation::id, reply_content> m_known_content;
                };
                
             } // content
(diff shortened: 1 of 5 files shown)


middleware/domain/include/domain/discovery/state.h
```diff
@@ -356,6 +356,68 @@ namespace casual
...
+                  //! adds the content and associate with the correlation.
+                  void add( const common::strong::correlation::id& correlation, const request_content& content);
+
+                  template< typename M>
+                  void add( const M& message) { add( message.correlation, message.content);}
+
+                  //! adds all cached resources to services and queues
+                  request_content complement( request_content&& content);
+
+                  //! @returns the "cached" requested resources associated withe the correlation intersected with the
```
ahhud: spelling `withe`

middleware/domain/source/discovery/handle.cpp
```diff
@@ -226,25 +218,24 @@ namespace casual
                      {
                         message::discovery::Reply reply;
                         reply.domain = common::domain::identity();
-                        reply.content = std::move( content);
 
                         for( auto& destination : discoveries)
                         {
+                           // filter based on the requested resources.
+                           reply.content = state.in_flight_cache.filter_reply( destination.correlation, content);
```
ahhud, line 225: Nice!

middleware/domain/source/discovery/handle.cpp
```diff
@@ -497,6 +485,8 @@ namespace casual
                      return;
                   }
 
+                  state.in_flight_cache.add( message);
```
ahhud: In case of a local lookup path in below statements, I think there is a obsolete cache registration never removed.

middleware/domain/source/discovery/state.cpp
```diff
@@ -228,19 +228,115 @@ namespace casual
...
+                     auto make_range = []( auto pair) { return range::make( pair.first, pair.second);};
+
+                     auto resources = make_range( m_correlation_to_resource.equal_range( correlation));
+
+                     result.reserve( resources.size());
+
+                     for( auto& pair : resources)
+                     {
+                        // clean upp total "cached" services
+                        if( auto found = algorithm::find( m_resource_count, pair.second); --found->second == 0)
```
ahhud, line 267: Personally I think this type of compound statement is somewhat hard to read and grasp. But that is just because I am not used to it. :-)

ahhud (changes requested):
Its really only the possible obsolete cache registration which needs addressing. Rest is more subjective. :-)

middleware/domain/source/discovery/handle.cpp
lazan: Yeah, I think you're right.

middleware/domain/source/discovery/state.cpp
lazan: You mean with _if init statement_ in general?

lazan: In some cases you can narrow the scope of a variable to the if scope with _init-statement_, which often is worth it. In this particular case it could be done with two nested if statements. Which is ok if it's one-liners, I think.

ahhud: If there is a very long init statement the actual condition appear at the very far right. 
Perhaps a different format is clearer. I.e 
if( init...;
     condition)
   statement;

But it is only a matter of taste.

ahhud (approved):

Issue #217, opened 2023-06-08
service - call inexplicably fails when starting domain under heavy load


AlbsAgren:
The callee shows no trace of ever having received the call.

Working theory: the service manager sends too multiple 'busy' replies for the same lookup request.

AlbsAgren:
When receiving a event::process::Exit the service manager will go through all pending lookups, regardless of wheter they had anything to do with the exiting process. This may cause the manger to send lookup replies that are semantically invalid, causing calls to "dissapear".

PR #219, opened 2023-06-09
service - avoid double busy lookup

AlbsAgren:
Before: When receiving a concurrent::Advertise, the service manager would retry all pending lookups, potentially leading to a situation where a caller would get double 'busy' replies.

Now: On receiving concurrent::Advertise we only retry pending lookups for services that have concurrent instances. Additionally we raise an error in the caller upon receiving multiple busy looup replies.

Addresses #217

### DIFF
--- a/middleware/common/source/service/call/context.cpp
+++ b/middleware/common/source/service/call/context.cpp
@@ -168,6 +168,9 @@ namespace casual
             target = service();
          }
 
+         if( target.state != decltype( target.state)::idle)
+            code::raise::error( code::casual::invalid_semantics, "unable to reserve instance of service '", target.service, "'");
+
          // Call the service
          {
             prepared.message.service = target.service;
--- a/middleware/service/source/manager/state.cpp
+++ b/middleware/service/source/manager/state.cpp
@@ -636,10 +636,15 @@ namespace casual
          // remove
          local::remove_services( *this, instance, message.services.remove);
 
-         // find all potentially pending.
+         // find all potentially pending that might be enabled by the new concurrent service(s)
          auto [ keep, remove] = algorithm::stable::partition( pending.lookups, [&]( auto& pending)
          {
-            if( auto found = algorithm::find( services, pending.request.requested))
+            auto is_requested_concurrent_service = [ &]( const auto& service)
+            {
+               return service.second.is_concurrent() && ! service.second.is_sequential() && service.first == pending.request.requested;
+            };
+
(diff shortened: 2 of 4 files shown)


middleware/common/source/service/call/context.cpp
```diff
@@ -168,6 +168,9 @@ namespace casual
             target = service();
          }
 
+         if( target.state != decltype( target.state)::idle)
+            code::raise::error( code::casual::invalid_semantics, "unable to reserve instance of service '", target.service, "'");
+
```
lazan, line 173: Is this some sort of sanity check?

middleware/service/source/manager/state.cpp
```diff
@@ -636,10 +636,15 @@ namespace casual
...
 
-         // find all potentially pending.
+         // find all potentially pending that might be enabled by the new service(s)
          auto [ keep, remove] = algorithm::stable::partition( pending.lookups, [&]( auto& pending)
          {
-            if( auto found = algorithm::find( services, pending.request.requested))
+            auto is_requested_concurrent_service = [ &]( const auto& service)
+            {
+               return service.second.is_concurrent() && service.first == pending.request.requested;
+            };
```
lazan, line 645: Hmm, is this totally safe? The service could be _sequential_ also, right? If so, couldn't we get some problems?
I think we need to check  `if( is-concurrent && ! is-sequential && service is the one we want) ` or something like: `service.second.is_only_concurrent()` ?

middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1047,5 +1047,50 @@ namespace casual
...
+            auto service = common::service::Lookup{ "local-service"}();
+            EXPECT_TRUE( service.service.name == "local-service");
+            EXPECT_TRUE( service.state == decltype( service.state)::idle);
+         }
+
+         // lookup 'local-service' again
+         common::service::Lookup lookup{ "local-service"};
+
+         // some unrelated concurrent services are advertised while our second lookup is pending
+         service::unittest::concurrent::advertise( { "some-remote-service", "some-other-remote-service"});
```
lazan, line 1069: Nice!

middleware/service/source/manager/state.cpp
AlbsAgren: Yes, I think you're right

lazan (approved):

PR #230, opened 2023-06-22
service: handle multiple busy lookups

### LINKED ISSUE #217: service - call inexplicably fails when starting domain under heavy load

AlbsAgren:
Before: we relied on the service manager only sending a single busy lookup reply before sending idle. The SM didnt always uphold this contract which would occasionally cause service calls to fail.

Now: we wait until we receive an idle reply irrespective of how many busy ones are sent beforehand.

Resolves #217

### DIFF
--- a/middleware/common/source/service/lookup.cpp
+++ b/middleware/common/source/service/lookup.cpp
@@ -35,7 +35,8 @@ namespace casual
 
          Lookup::~Lookup()
          {
-            if( m_correlation)
+            // we have to discard if we're still pending with the service manager
+            if( ! m_reply || m_reply->busy())
                lookup::discard( m_correlation);
          }
 
@@ -69,25 +70,14 @@ namespace casual
             swap( lhs.m_correlation, rhs.m_correlation);
          }
 
-
          void Lookup::update( Reply&& reply)
          {
             log::line( verbose::log, "reply: ", reply);
 
-            switch( reply.state)
-            {
-               case State::idle:
-                  m_correlation = {};
-                  m_reply = std::move( reply);
-                  break;
-               case State::absent:
-                  m_correlation = {};
-                  code::raise::error( code::xatmi::no_entry, "lookup: ", *this);
-                  break;
-               case State::busy:
-                  m_reply = std::move( reply);
-                  break;
-            }
+            m_reply = std::move( reply);
+
+            if( m_reply->state == State::absent)
+               code::raise::error( code::xatmi::no_entry, "lookup: ", *this);
          }
       } // detail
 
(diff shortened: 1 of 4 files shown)


middleware/common/source/service/lookup.cpp
```diff
@@ -69,7 +69,6 @@ namespace casual
             swap( lhs.m_correlation, rhs.m_correlation);
          }
 
-
          void Lookup::update( Reply&& reply)
```
AlbsAgren, line 73: I decided against removing this. Resetting m_correlation is used to indicate that we dont need to discard the pending lookup so it felt a bit tricky breaking this out into a free function.

middleware/common/source/service/lookup.cpp
```diff
@@ -122,11 +121,18 @@ namespace casual
       {
          Trace trace{ "common::service::Lookup::operator()"};
```
AlbsAgren, line 113: The non-blocking ditto of this already appears to wait until it gets a 'non-busy' reply, so it should work as-is.

middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1092,5 +1092,55 @@ namespace casual
...
+         // lookup 'local-service' again
+         common::service::Lookup lookup{ "local-service"};
+
+         // meanwhile, some process dies
+         {
+            common::message::event::process::Exit event;
+            event.state.pid = common::strong::process::id{ 42};
+            event.state.reason = decltype( event.state.reason)::exited;
+            common::communication::device::blocking::send( common::communication::instance::outbound::service::manager::device(), event);
+         }
```
lazan, line 1119: Nice

middleware/common/source/service/lookup.cpp
```diff
@@ -122,11 +121,18 @@ namespace casual
...
          {
-            Reply result;
-            communication::device::blocking::receive( communication::ipc::inbound::device(), result, m_correlation);
-            update( std::move( result));
+            // wait until service is idle
+            // the second lookup::Reply should always be idle, but the service manager
+            // is currently not completely reliable in that regard
+            // TODO: rework this after fixing service manager
+            while( m_reply->busy())
+               update( communication::ipc::receive< Reply>());
```
lazan, line 126: NIce

lazan (approved):

lazan (approved):
Feels a lot simpler.

ahhud (approved):

PR #235, opened 2023-07-18
Feature/1.7/refactoring

lazan:
A lot of refactoring stuff. Mostly domain task stuff.


OlofGullnas:
Noticed that this PR requires updated casual/casual-thrdparty/master in addition to an (obvious) update of repository casual/casual-make.

Not really unexpected ;-)

lazan:
> Noticed that this PR requires updated casual/casual-thrdparty/master in addition to an (obvious) update of repository casual/casual-make.
> 
> Not really unexpected ;-)

Yeah 😄  Do you think that should have been described in the PR or perhaps in the commit message?

OlofGullnas:
Yes. I think it would have been nice to mention the need for an updated third-party in the commit message or PR. I checked the bug report you made aqainst the pugixml XML library. As I understand it there is a newer version where the more modern C++ compatibility  issue was resolved. Should we change to a latest version in casual-thirdparty?

middleware/buffer/include/casual/buffer/internal/field/string.h
```diff
@@ -217,8 +217,8 @@ namespace casual
                      }
                      
                      // to string
-                     template< typename T>
-                     auto operator() ( T value, stream::Output::view_type destination, char padding) -> std::enable_if_t< std::is_integral< T>::value>
+                     template< std::integral T>
```
AlbsAgren, line 220: Woo, concepts!

PR #245, opened 2023-08-22
Patch/1.6/nginx problem

lazan:


AlbsAgren (approved):
I'm still not entirely sure how NGX_AGAIN is supposed to work, but if it passes our tests I guess we just have to keep an eye on this.

lazan:
Merged manually

PR #246, opened 2023-08-22
configuration: prevent global contract setting being default overridden

### LINKED ISSUE #244: configuration - default values override user specified values when using multiple config files
Consider the following config files:

global.yaml
```
domain:
  name: A

  global:
    service:
      execution:
        timeout:
          duration: 45
          contract: "terminate"
```

other.yaml
```
domain:
  servers:
```

If casual is started with `casual domain -b global.yaml other.yaml` the default values for services will be a timeout duration of 45s but the default contract will be "linger", even though other.yaml doesnt explicitly override this value.

Observed for the "contract" option in the global service config, might be present elsewhere.

AlbsAgren:
Before:
A specified global execution.timeout.contract setting would be overriden by the default value when using multiple configuration files.

Now:
The setting is only overridden by the rightmost config file where it has been explicitly set.

Resolves #244

### DIFF
--- a/middleware/configuration/unittest/source/test_model.cpp
+++ b/middleware/configuration/unittest/source/test_model.cpp
@@ -685,6 +685,65 @@ domain:
          EXPECT_TRUE( A == result) << CASUAL_NAMED_VALUE( A) << "\n" << CASUAL_NAMED_VALUE( result);
       }
 
+
+      TEST( configuration_model_set_operations, A_union_B__partial_override_of_A_in_B)
+      {
+         common::unittest::Trace trace;
+
+         constexpr auto configuration = R"(
+domain:
+   name: A
+   global:
+      service:
+         execution:
+            timeout:
+               duration: 90
+               contract: terminate
+   servers:
+      - alias: a
+        path: a
+)";
+
+         // note: we do not override contract
+         constexpr auto add = R"(
+domain:
+   name: B
+   global:
+      service:
+         execution:
+            timeout:
+               duration: 45
+   servers:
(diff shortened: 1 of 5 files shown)


middleware/configuration/unittest/source/test_model.cpp
```diff
@@ -685,6 +685,65 @@ namespace casual
...
+
+         auto A = local::configuration( configuration);
+         auto B = local::configuration( add);
+         auto result = local::configuration( expected);
+
+         A = set_union( A, std::move( B));
+
+         EXPECT_TRUE( A == result) << CASUAL_NAMED_VALUE( A) << "\n" << CASUAL_NAMED_VALUE( result);
+      }
+
```
lazan, line 746: Easy to understand unittest.

middleware/service/source/manager/handle.cpp
```diff
@@ -162,7 +163,7 @@ namespace casual
                // send event, at least domain-manager want's to know...
                common::message::event::process::Assassination event{ common::process::handle()};
                event.target = entry.target;
-               event.contract = entry.service->timeout.contract;
+               event.contract = entry.service->timeout.contract.value_or( common::service::execution::timeout::contract::Type::linger);
```
lazan: Do we take global into account during transformation of the configuration? It seems like there should be code changes for that? I mean, if duration is specified but not contract for a service, then the global contract should be used.

lazan (approved):

lazan:
Merged manually

PR #249, opened 2023-08-31
Patch/1.6/service forward bug

lazan:


AlbsAgren (approved):
Looks good, I think the message counter has the potential to be quite useful going forward

PR #251, opened 2023-08-31
service - improve handling of service timeout

AlbsAgren:
* We now take the global config into account when calculating the timeout duration for services that are featured in the configuration
* Timeout contract is now displayed correctly when listing services with the cli
* A service with no timeout duration is now indicated by '-' rather than 0.00 when listing services with the cli


lazan (approved):
Looking good.

PR #263, opened 2023-10-06
queue - send reply on failed dequeues

### LINKED ISSUE #262: queue - a full queuebase may cause forward to hang
The query to dequeue a message may fail with `[sql:database or disk is full] database or disk is full: database or disk is full failed with dequeue request from queue:` in which case the forward does not recieve a reply to its dequeue request and is left waiting forever.

AlbsAgren:
Before:
If sqlite threw an exception before we had dequeued a message, no dequeue_reply would be sent.

Now:
We make sure to always send a reply so that the caller can react appropriately.

Resolves #262

### DIFF
--- a/middleware/queue/source/group/handle.cpp
+++ b/middleware/queue/source/group/handle.cpp
@@ -243,9 +243,8 @@ namespace casual
                      Trace trace{ "queue::handle::dequeue::Request::handle"};
                      log::line( verbose::log, "message: ", message);
 
-                     auto now = platform::time::clock::type::now();
-
-                     auto reply = state.queuebase.dequeue( message, now);
+                     // send an empty reply on error TODO: is this the api we want?
+                     auto reply = common::message::reverse::type( message);
                      reply.correlation = message.correlation;
 
                      // make sure we always send reply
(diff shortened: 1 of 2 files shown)


middleware/queue/source/group/handle.cpp
```diff
@@ -243,9 +243,8 @@ namespace casual
                      Trace trace{ "queue::handle::dequeue::Request::handle"};
                      log::line( verbose::log, "message: ", message);
 
-                     auto now = platform::time::clock::type::now();
-
-                     auto reply = state.queuebase.dequeue( message, now);
+                     // send an empty reply on error TODO: is this the api we want?
```
ahhud, line 246: How does the _forward_ act on an empty reply? And is it detected as an error?

AlbsAgren: It logs an error and rolls back the transaction, restarting the 'flow' when it gets the rollback_reply. At least that's my interpretation: https://github.com/casualcore/casual/blob/patch/1.6/main/middleware/queue/source/forward/handle.cpp#L490

lazan (approved):
Nice!

middleware/queue/source/group/handle.cpp
ahhud: Ok

ahhud (approved):

lazan:
Merged manually

PR #265, opened 2023-10-10
gateway: pending 'calls' could be sent after a rollback

### LINKED ISSUE #261: timeout could render premature rollback -> stale transaction
* A calls B 
* B is busy -> the call is pending in B inbound
* A timeouts
* A get assassinate
* TM in A sends rollback
* TM in B has no RM associated -> read-only
* B gets idle -> the pending call is sent
* The call associate the transaction with B:s TM
* No one will do anything with the transaction -> the transaction is stale.

Some mitigation have been done to 1.6, but there still problem.

lazan:
Pending request (call/conversation/de/en-queue) could be sent after a rollback request (that was triggered by a timeout).

This is really a bigger problem, and this fix does not fix it to 100%.

Somewhat resolves #261

### DIFF
--- a/middleware/gateway/include/gateway/group/inbound/state.h
+++ b/middleware/gateway/include/gateway/group/inbound/state.h
@@ -68,11 +68,10 @@ namespace casual
                }
                
                template< typename M>
-               inline auto add( M&& message)
+               inline auto add( const M& message)
                {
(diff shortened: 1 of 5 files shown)


AlbsAgren (approved):
It's unfortunate that we don't have a "100%" fix, but i think this is a good pragmatic solution for now.

lazan:
Merged manually

PR #268, opened 2023-10-16
xatmi: add function to reset execution_id from public api

### LINKED ISSUE #239: expose a function to reset the execution_id in the public api

albinsjolund:
Resets the execution_id and returns the new id

Resolves #239

### DIFF
--- a/middleware/xatmi/source/xatmi/extended.cpp
+++ b/middleware/xatmi/source/xatmi/extended.cpp
@@ -144,6 +144,12 @@ const uuid_t* casual_execution_id_get()
    return &casual::common::execution::id().underlying().get();
 }
 
+const uuid_t* casual_execution_id_reset()
+{
+   casual::common::execution::reset();
+   return &casual::common::execution::id().underlying().get();
+}
+
 void casual_instance_browse_services( casual_instance_browse_callback callback, void* context)
 {
    using namespace casual;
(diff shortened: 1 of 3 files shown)


middleware/xatmi/source/xatmi/extended.cpp
```diff
@@ -144,6 +144,12 @@ const uuid_t* casual_execution_id_get()
    return &casual::common::execution::id().underlying().get();
 }
 
+const uuid_t* casual_execution_id_reset()
+{
+   casual::common::execution::reset();
```
AlbsAgren, line 149: I have a vague memory that this function mucks around with environment variables in a way that is susceptible to memory leaks. Looks like it's mentioned in 9dd06ca. Not sure how applicable that issue is here, thoughts @lazan?

lazan: Yeah, exactly. This will leak memory on linux. But, I think the usage of _set environment_ in `execution::reset()` is a bug we need to fix regardless of exposing c-api to reset the execution-id. 
I'll create a separate issue regarding that (we might have one)

lazan (approved):
Nice.

middleware/xatmi/source/xatmi/extended.cpp
lazan: Created #275.

ahhud (approved):

AlbsAgren (approved):

lazan:
Merged manually

PR #274, opened 2023-10-29
Patch/1.6/multiple bug fixes

lazan:
The transaction cli commit is quite large, but I couldn't split it in compilable chunks. (I think every commit should be compilable, but smaller commits might be a better trade of).

### DIFF
--- a/middleware/queue/source/group/queuebase/statement.cpp
+++ b/middleware/queue/source/group/queuebase/statement.cpp
@@ -51,12 +51,13 @@ FROM
    message
 WHERE queue = :queue AND state = 2 AND available < :available ORDER BY timestamp ASC, available ASC LIMIT 1; )");
 
+            // for dequeue with an explicit id we ignore _available_
             result.dequeue.first_id = connection.precompile( R"(
 SELECT
    ROWID, id, properties, reply, redelivered, type, available, timestamp, payload
 FROM 
    message
-WHERE id = :id AND queue = :queue AND state = 2 AND available < :available; )");
+WHERE id = :id AND queue = :queue AND state = 2; )");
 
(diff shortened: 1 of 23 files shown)


middleware/queue/source/group/queuebase/statement.cpp
```diff
@@ -51,12 +51,13 @@ FROM
...
 WHERE queue = :queue AND state = 2 AND available < :available ORDER BY timestamp ASC, available ASC LIMIT 1; )");
 
+            // for dequeue with an explicit id we ignore _available_
             result.dequeue.first_id = connection.precompile( R"(
 SELECT
    ROWID, id, properties, reply, redelivered, type, available, timestamp, payload
 FROM 
    message
-WHERE id = :id AND queue = :queue AND state = 2 AND available < :available; )");
+WHERE id = :id AND queue = :queue AND state = 2; )");
```
ahhud, line 60: I understand that you might want to dequeue from cli not taken available into account.
But isn't there usecases when you actual _not_ want the message to be available directly in some kind of retry situation? I know that in those situations you normally dequeue without an id but isn't it possible that you want to be able to do that?

lazan: Well, we've decided that if you provide id for the message, we ignore the potentially available. This to suport needs we know about.

I can't see any use-cases except some sort of pull timer (most expensive timer ever 😄)

ahhud: Well, my point was really that there can be different requirements and behavior on an _admin cli_ than on the external api. But we address that when we get there.

ahhud (approved):

PR #276, opened 2023-11-01
Patch/1.6/rm proxy metric

lazan:


middleware/transaction/source/manager/handle.cpp
```diff
@@ -593,7 +593,7 @@ namespace casual
                               {
                                  // We got a pending request for this resource, let's oblige
                                  if( state.multiplex.send( instance.process.ipc, std::move( request.complete)))
-                                    instance.reserve();
+                                    instance.reserve( request.created);
```
lazan, line 596: A bug that I discovered with the total cli-unittest.

ahhud (approved):

PR #278, opened 2023-11-01
domain: handle restarts attribute correct

### LINKED ISSUE #277: Process restarts attribute doesn't calculates correctly
Casual doesn´t keep track of restarts attribute for servers and executables at all causing attribute always showing 0 in cli etc.

ahhud:
restarts attribute for servers and executeble are incremented when they are restarted for any reason.

Resolves #277

### DIFF
--- a/middleware/domain/include/domain/manager/state.h
+++ b/middleware/domain/include/domain/manager/state.h
@@ -108,8 +108,7 @@ namespace casual
 
             bool restart = false;
 
-            //! Number of instances that has been restarted
-            platform::size::type restarts = 0;
+            platform::size::type initiated_restarts = 0;
 
             inline friend bool operator < ( const Process& l, const Process& r) { return l.id < r.id;}
             inline friend bool operator == ( const Process& lhs, const std::string& alias) { return lhs.alias == alias;}
@@ -124,7 +123,7 @@ namespace casual
                CASUAL_SERIALIZE( memberships);
                CASUAL_SERIALIZE( environment);
                CASUAL_SERIALIZE( restart);
-               CASUAL_SERIALIZE( restarts);
+               CASUAL_SERIALIZE( initiated_restarts);
             )
          };
 
--- a/middleware/domain/source/manager/admin/cli.cpp
+++ b/middleware/domain/source/manager/admin/cli.cpp
@@ -348,7 +348,7 @@ namespace casual
    restart;
       if instances should be restarted if exited
    #r:
-      number of times instances has been restarted
+      number of times instances has been initiated for a restart
    path:
       the path to the binary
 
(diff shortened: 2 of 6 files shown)


middleware/domain/source/manager/handle.cpp
```diff
@@ -601,10 +601,19 @@ namespace casual
                            else
                               log::line( log::category::information, "process exited, alias: ", alias, ", details: ", message.state);
 
+                           auto scale = [ &state]( auto& executable)
+                           {
+                              auto spawnable = executable->spawnable();
+                              handle::scale::instances( state, *executable);
+                              if( spawnable)
+                                 executable->restarts++;
+                           };
```
lazan: * What if scale::instances fails? 
* What if there are more than one _spawnable_? 

It's really unfortunate that "we" picked _restarts_... It gets way more complicated then one would like...

Should we rethink this?

middleware/domain/unittest/source/manager/test_manager.cpp
```diff
@@ -1672,5 +1672,76 @@ namespace casual
...
+
+         constexpr auto sleep =  R"(
+domain:
+   name: sleep
+   executables:
+      -  path: sleep
+         alias: sleep
+         arguments: [60]
+         instances: 1
+         restart: true
```
lazan, line 1688: Nice. Clever with sleep, even if we have a bug it will only "hang" 60s.

middleware/domain/source/manager/handle.cpp
lazan: What if we keep track of deaths/exits? And then calculated the restarts for the cli? No that wont work...

lazan: I've been looking for a less "complex" way... The only thing that seems reasonable is to increment `restarts` in  `state::Server::remove( common::strong::process::id pid)` and  `state::Executable::remove( strong::process::id pid)`.  That's when we go from `running` to  `scale_out` via _restart_ .  The meaning of _restarts_ becomes more like _scheduled_restarts_, but to me it feels more sane. 
What do you think @ahhud

middleware/domain/include/domain/manager/state.h
```diff
@@ -124,7 +123,7 @@ namespace casual
                CASUAL_SERIALIZE( memberships);
                CASUAL_SERIALIZE( environment);
                CASUAL_SERIALIZE( restart);
-               CASUAL_SERIALIZE( restarts);
+               CASUAL_SERIALIZE( initiated_restarts);
```
lazan, line 126: Good name.

middleware/domain/source/manager/admin/cli.cpp
```diff
@@ -348,7 +348,7 @@ namespace casual
    restart;
       if instances should be restarted if exited
    #r:
-      number of times instances has been restarted
+      number of times instances has been initiated for a restart
```
lazan, line 351: Nice!

lazan (approved):
Feels like this makes sense. Nice.

PR #280, opened 2023-11-06
cli: add precondition casual log can't be tied to stdout

### LINKED ISSUE #259: cli - casual pipe breaks when logging to standard out
When logging to stdout, any logs produced during the execution of one casual command will be piped to the next one. This breaks the following command if it tries to read from stdin.

lazan:
This patch checks if casual log is tied to stdout and if so, we exit the cli and print the violated precondition to stderr.

Resolves #259

### DIFF
--- a/middleware/administration/source/main.cpp
+++ b/middleware/administration/source/main.cpp
@@ -8,20 +8,60 @@
 #include "casual/administration/cli.h"
 
 #include "common/exception/guard.h"
+#include "common/environment.h"
 
-#include <iostream>
-
+// TODO remove in 1.7
+#include <sys/stat.h>
 
 namespace casual
-{  
+{
+   using namespace common;
+
    namespace administration
    {
       namespace local
       {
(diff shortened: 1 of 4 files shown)


middleware/administration/source/main.cpp
```diff
@@ -8,20 +8,31 @@
...
    namespace administration
    {
       namespace local
       {
          namespace
          {
+            void validate_preconditions()
+            {
+               if( std::filesystem::equivalent( environment::log::path(), "/dev/stdout"))
+                  code::raise::error( code::casual::preconditions, "causal log can't be tied stdout when using cli");
```
AlbsAgren: Typo in 'causal'

AlbsAgren (approved):

middleware/administration/source/main.cpp
lazan: good catch.

PR #284, opened 2023-11-07
xatmi - comply to specification signal handling

### LINKED ISSUE #271: signal handling
There are some potential issues with signal handling that should be investigated.
- are interrupts propagated to the user though TPGOTSIG?
- do we comply to the spec regarding the TPSIGRSTRT flag?
- is there a need for a TPSIGRSTRT equivalent in the queue-api?

AlbsAgren:
Before: We did not correctly handle signals according to the specification for the TPSIGRSTRT flag.
Nor did we set tperrno to TPGOTSIG if a call failed due to receiving a signal from the OS.

Now: For the XATMI functions where TPSIGRSTRT is valid, we now block incoming signals for the duration of the call. If TPSIGRSTRT is not set and a call fails due to a signal, we now set tperrno to TPGOTSIG.

Resolves #271

### DIFF
--- a/middleware/xatmi/include/casual/xatmi/internal/signal.h
+++ b/middleware/xatmi/include/casual/xatmi/internal/signal.h
@@ -1,0 +1,26 @@
+//!
+//! Copyright (c) 2018, The casual project
+//!
+//! This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+//!
+
+#pragma once
+
+#include "common/signal.h"
+
+#include <optional>
+
+
+namespace casual::xatmi::internal::signal
+{
+    template< typename Flags>
+    inline std::optional< casual::common::signal::thread::scope::Block> maybe_block( Flags flags)
+    {
+        using enum_type = typename Flags::enum_type;
+
+        if( flags.exist( enum_type::signal_restart))
+            return casual::common::signal::thread::scope::Block{};
+
+        return {};
+    }
+} // casual::xatmi::internal::signal
--- a/middleware/xatmi/source/xatmi/internal/code.cpp
+++ b/middleware/xatmi/source/xatmi/internal/code.cpp
@@ -10,6 +10,7 @@
 #include "common/exception/capture.h"
 #include "common/code/category.h"
 #include "common/log.h"
+#include "common/code/signal.h"
 
 namespace casual
 {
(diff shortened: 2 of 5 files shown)


middleware/xatmi/source/xatmi/internal/code.cpp
```diff
@@ -28,6 +29,9 @@ namespace casual
                if( code == common::code::casual::invalid_argument)
                   return common::code::xatmi::argument;
 
+               if( common::code::is::category< common::code::signal>( code))
+                  return common::code::xatmi::signal;
+
```
lazan, line 34: We can't get a `code::casual::interrupted` here?

middleware/xatmi/include/casual/xatmi/internal/signal.h
```diff
@@ -0,0 +1,26 @@
...
+namespace casual::xatmi::internal::signal
+{
+    template< typename Flags>
+    inline std::optional< casual::common::signal::thread::scope::Block> maybe_block( Flags flags)
+    {
+        using enum_type = typename Flags::enum_type;
+
+        if( flags.exist( enum_type::signal_restart))
+        return casual::common::signal::thread::scope::Block{};
+
```
lazan, line 23: indentation?

test/unittest/source/xatmi/test_call.cpp
```diff
@@ -622,6 +623,108 @@ namespace casual
...
+         EXPECT_TRUE( tperrno == 0) << "tperrno: " << tperrnostring( tperrno);
+
+         signal::send( process::id(), code::signal::alarm);
+
+         EXPECT_TRUE( tpgetrply( &desc, &buffer, &len, TPSIGRSTRT) == 0);
+         EXPECT_TRUE( tperrno == 0) << "tperrno: " << tperrnostring( tperrno);
+
+         signal::clear();
+      }
+
```
lazan, line 727: Nice unittests!

middleware/xatmi/include/casual/xatmi/internal/signal.h
```diff
@@ -0,0 +1,26 @@
...
+        return {};
+    }
```
lazan, line 25: This, with an optional, feels like a good idiom. Easy to reason about.

middleware/xatmi/source/xatmi/internal/code.cpp
AlbsAgren: From what I've been able to tell, where we might get `code::casual::interrupted`, we check for it and then call `signal::dispatch()` to handle the signal or raise a `code::signal`. Though there might be something I'm not seeing.

lazan (approved):

PR #285, opened 2023-11-07
service: perform discovery on external_discovery

### LINKED ISSUE #269: service - lookup requests with external_discovery do not trigger a discovery if no instances of service are known
The effect of this may be that requests from another domain get TPENOENT even if the service has become available in another connected domain. When the calling domain gets the reply, it will unadvertise the service and perform a new discovery the next time the service is called, so any subsequent requests will be successful (not including ones that have already been sent).

One caveat is that changing this could generate a large amount of discovery requests, with unknown effects on performance.

ahhud:
Before: No discovery was made for external_discovery requester if service was known by service-manager but no routing existed and tpenoent was returned.

After: A new discovery was made and the service was found.

Resolves #269

### DIFF
--- a/test/unittest/source/test_gateway.cpp
+++ b/test/unittest/source/test_gateway.cpp
@@ -3363,6 +3363,116 @@ domain:
          
       }
 
+
+
+      TEST( test_gateway, domain_A_to_GW_to_B__perform_discovery_in_GW_if_service_unknown)
+      {
+         common::unittest::Trace trace;
+
+         // sink child signals
+         signal::callback::registration< code::signal::child>( [](){});
+
+         // domain containing example services scaled down to 0 instances
+         constexpr auto B = R"(
+domain:
+   name: B
+
+   servers:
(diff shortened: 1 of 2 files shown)


AlbsAgren (approved):

test/unittest/source/test_gateway.cpp
```diff
@@ -3363,6 +3363,116 @@ namespace casual
...
+            casual::service::unittest::fetch::until( casual::service::unittest::fetch::predicate::instances( "casual/example/echo", 1));
+         }
+
+         // domain a: routing to service via gw exists
+         // domain gw: example service is known by service-manager after previous call, but no routing to domain b exists for service
+         // the call trigger a new discovery for service in gw instead of tpenoent
+         a.activate();
+         local::call( "casual/example/echo", code::xatmi::ok);
+
+      }
```
lazan, line 3471: Pretty good unittest! The last comments ties the whole thing together.

lazan (approved):

Issue #302, opened 2024-01-09
casual_field_copy_memory changes buffer type to NULL if source is empty


AlbsAgren:
One would presumably expect to end up with an empty buffer of the same type? Alternatively CASUAL_FIELD_INVALID_ARGUMENT if we dont consider it valid calling the function with an empty source.

```cpp
TEST( buffer_field, test_buffer_type)
{
   auto buffer = tpalloc( CASUAL_FIELD, nullptr, 512);
   ASSERT_TRUE( buffer != nullptr);

   {
      std::array< char, 8> type;
      auto size = tptypes( buffer, type.data(), nullptr);

      ASSERT_TRUE( size != -1);
      ASSERT_TRUE( std::string{ type.data()} == CASUAL_FIELD) << type.data();
   }

   std::vector< char> bytes{};

   ASSERT_TRUE( casual_field_copy_memory( &buffer, bytes.data(), bytes.size()) == CASUAL_FIELD_SUCCESS);

   {
      std::array< char, 8> type;
      auto size = tptypes( buffer, type.data(), nullptr);

      ASSERT_TRUE( size != -1);
      // currently the buffer has type NULL here
      ASSERT_TRUE( std::string{ type.data()} == CASUAL_FIELD) << type.data();
   }
}
```

PR #306, opened 2024-01-25
Feature/1.7/buffer fixes

KristianIvarsson:
Fixes for issue #302 and some other minor return code fixes

### DIFF
--- a/middleware/buffer/source/field.cpp
+++ b/middleware/buffer/source/field.cpp
@@ -97,7 +97,7 @@ namespace casual
                   }
 
                   if( cursor > end)
-                     common::code::raise::error( common::code::xatmi::argument, "buffer is comprised");
+                     throw std::invalid_argument{ "buffer is comprised"};
                }
 
                // invoked when a service is invoked with the "buffer from the wire".
@@ -170,7 +170,7 @@ namespace casual
                // called on service invocation
                common::buffer::handle::mutate::type adopt( common::buffer::Payload&& payload, common::buffer::handle::type* inbound)
                {
-                  Trace trace{ "buffer::field::Allocator::adopt"};
+                  //Trace trace{ "buffer::field::Allocator::adopt"};
                   auto& buffer = allocator_base::emplace_back( std::move( payload), inbound);
                   return buffer.handle();
                }
@@ -215,7 +215,7 @@ namespace casual
 
             static_assert( common::buffer::pool::is::adoptable< Allocator>, "not adoptable"); 
 
-         } // <unnamed>
+         } //
       } // local
 
       namespace
@@ -247,7 +247,7 @@ namespace casual
                   const auto error = common::exception::capture();
 
                   if( error.code() == common::code::xatmi::argument)
-                     return CASUAL_FIELD_INVALID_HANDLE;
+                     return CASUAL_FIELD_INVALID_HANDLE; 
 
                   common::stream::write( std::cerr, "error: ", error, '\n');
                   
@@ -318,7 +318,7 @@ namespace casual
             template<typename... A>
             int data( char** handle, const item_type id, const int type, A&&... arguments)
             {
-               //const trace trace( "field::add::data");
+               //Trace trace( "field::add::data");
 
                if( type != (id / CASUAL_FIELD_TYPE_BASE))
                {
@@ -389,7 +389,7 @@ namespace casual
             template<typename... A>
             int data( const char* const handle, const item_type id, size_type occurrence, const int type, A&&... arguments)
             {
-               //const trace trace( "field::get::data");
+               //Trace trace( "field::get::data");
 
                if( type != (id / CASUAL_FIELD_TYPE_BASE))
                   return CASUAL_FIELD_INVALID_ARGUMENT;
@@ -455,7 +455,7 @@ namespace casual
 
             int data( const char* const handle, const item_type id, size_type occurrence)
             {
-               //const trace trace( "field::cut::data");
+               //Trace trace( "field::cut::data");
 
                if( ! (id > CASUAL_FIELD_NO_ID))
                   return CASUAL_FIELD_INVALID_ARGUMENT;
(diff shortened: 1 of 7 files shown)


middleware/buffer/source/field.cpp
```diff
@@ -97,7 +97,7 @@ namespace casual
                   }
 
                   if( cursor > end)
-                     common::code::raise::error( common::code::xatmi::argument, "buffer is comprised");
+                     throw std::invalid_argument{ "buffer is comprised"};
                }
```
OlofGullnas, line 96: Is "comprised" the correct word? The word "comprise" with past sense "comprised" has the meaning "consists of" eller "be made up of". Is it the word "compromised" that shold be used?  Or perhaps even "corrupt".

middleware/buffer/source/field.cpp
```diff
@@ -170,7 +170,7 @@ namespace casual
                // called on service invocation
                common::buffer::handle::mutate::type adopt( common::buffer::Payload&& payload, common::buffer::handle::type* inbound)
                {
-                  Trace trace{ "buffer::field::Allocator::adopt"};
+                  //Trace trace{ "buffer::field::Allocator::adopt"};
                   auto& buffer = allocator_base::emplace_back( std::move( payload), inbound);
```
OlofGullnas, line 180: Should Trace be commented? If this is intentional, should the line be removed instead?

<edit>
I noticed that there are several "//Trace ...." later in the file, so perhaps this is according to "style" in this file. Statements provided in commented form for easy turn on in case needed, but commented because they produces to much "noise" if turned on under normal circumstances.

middleware/buffer/source/field.cpp
```diff
@@ -247,7 +247,7 @@ namespace casual
                   const auto error = common::exception::capture();
 
                   if( error.code() == common::code::xatmi::argument)
-                     return CASUAL_FIELD_INVALID_HANDLE;
+                     return CASUAL_FIELD_INVALID_HANDLE;
```
OlofGullnas, line 257: Remove space at end of line? reduces noise in diff...

middleware/buffer/source/octet.cpp
```diff
@@ -98,8 +100,12 @@ namespace casual
...
                catch( ...)
                {
-                  if( common::exception::capture().code() == common::code::xatmi::argument)
-                     return CASUAL_OCTET_INVALID_HANDLE;
+                  const auto error = common::exception::capture();
+
+                  if( error.code() == common::code::xatmi::argument)
+                     return CASUAL_OCTET_INVALID_HANDLE; 
+
+                  common::stream::write( std::cerr, "error: ", error, '\n');
```
OlofGullnas, line 100: Shall this be logged to log file or go to std::cerr? (Or possibly both?)

Noted that this construct (write to stderr) occurs in several places.

middleware/buffer/source/field.cpp
KristianIvarsson: This error message have been there for many years and in practice it isn't even used for logging or anything so it can even be removed entierly

middleware/buffer/source/field.cpp
KristianIvarsson: I guess your analysis is correct

middleware/buffer/source/octet.cpp
KristianIvarsson: This was more a way to make the code/behaviour consistent with other C-buffers and the question is not just valid for this buffer type

middleware/buffer/source/octet.cpp
```diff
@@ -173,13 +173,12 @@ namespace casual
 
             }
 
-         } // <unnamed>
+         } //
```
lazan, line 167: I think the `<unnamed>` gives information. Not just that someone has forgotten to comment the end of a namespace. But it's not that important.

middleware/buffer/source/order.cpp
```diff
@@ -162,8 +163,12 @@ namespace casual
...
                   catch( ...)
                   {
-                     if( common::exception::capture().code() == common::code::xatmi::argument)
-                        return CASUAL_ORDER_INVALID_HANDLE;
+                     const auto error = common::exception::capture();
+
+                     if( error.code() == common::code::xatmi::argument)
+                        return CASUAL_ORDER_INVALID_HANDLE; 
+
+                     common::stream::write( std::cerr, "error: ", error, '\n');
```
lazan: Why do we write to `std::cerr` this could happen not only in CLI context I presume?

middleware/buffer/source/string.cpp
```diff
@@ -129,14 +129,22 @@ namespace casual
...
                catch( ...)
                {
-                  if( common::exception::capture().code() == common::code::xatmi::argument)
-                     return CASUAL_STRING_INVALID_HANDLE;
+                  const auto error = common::exception::capture();
+
+                  if( error.code() == common::code::xatmi::argument)
+                     return CASUAL_STRING_INVALID_HANDLE; 
+
+                  common::stream::write( std::cerr, "error: ", error, '\n');
```
lazan: Same here...

lazan (commented):
Why log to `std::cerr`?

middleware/buffer/source/order.cpp
KristianIvarsson: It's been like that for ages in CASUAL_FIELD, so it was just to make it coherent, but we can do whatever appropriate

middleware/buffer/source/string.cpp
KristianIvarsson: Shall we log it into casual-log instead and if so, what category ?

lazan: We've got the `common::log::error( <code>, <info>...)`. Logs to error category.  But what _code_ should you use? Check if you think you find something sufficient in `code::casual::???`. If not, maybe a `code::buffer` should be created?

lazan: If you create one, I think it should be local to _buffer_. Should it not?

lazan: Or, of course, you could add _codes_ in `code::casual` if you want.

lazan (approved):

lazan:
Merged manually

PR #313, opened 2024-02-13
queue: clear available when moving message to error queue

### LINKED ISSUE #312: Messages not immediately available on error queue if queue has retry delay
When messages are moved to error queue due to a rollback, they are not available for consumption until the retry delay has elapsed

partickl:
When a message is moved to the error queue, due to rollback, the 'available' field is now cleared so that the message can be consumed from the error queue immediately

Resolves #312


lazan (approved):
Looks nice. Easy test-case to understand.

lazan:
Merged manually

lazan:
Merged manually

PR #314, opened 2024-02-15
ipc: fix discard while message is only partially received

### LINKED ISSUE #311: Error if tpcancel is called while in the midst of recieving a large reply
If tpcancel is called while the caller has recieved parts of the reply, only the already received parts will be discarded, leading to the following error when the subsequent parts are received.

`[casual:internal-out-of-bounds] added transport has too large offset: 162880 - current offset: 0: internal-out-of-bounds`

AlbsAgren:
Before: We cleared the partially complete message from the device cache when being told to discard. This caused an out-of-bounds error when subsequent parts of the message arrived.
Now: We wait til the entire message has a arrived before clearing it from the cache.

Resolves #311

The unittest could perhaps be done in a more simple way, open to suggestions.

### DIFF
--- a/middleware/common/include/common/communication/device.h
+++ b/middleware/common/include/common/communication/device.h
@@ -259,11 +259,8 @@ namespace casual
 
             auto complete = algorithm::find_if( m_cache, [&]( auto& complete){ return complete.correlation() == correlation;});
 
-            if( complete)
+            if( complete && complete->complete())
             {
-               if( ! complete->complete())
-                  m_discarded.push_back( correlation);
-
                m_cache.erase( complete.begin());
             }
             else
@@ -402,7 +399,7 @@ namespace casual
                      if( discard( *found))
                         m_cache.erase( std::begin( found));
 
-                     // Try to find a massage that matches the predicate
+                     // Try to find a message that matches the predicate
                      found = algorithm::find_if( m_cache, predicate);
                   }
 
--- a/middleware/common/unittest/source/communication/test_ipc.cpp
+++ b/middleware/common/unittest/source/communication/test_ipc.cpp
@@ -223,7 +223,6 @@ namespace casual
          EXPECT_TRUE( ( algorithm::equal( receive_message.payload, send_message.payload)));
       }
 
-
       TEST( common_communication_ipc, send_receive__1__2x_transport_size__expect_exactly_2_transport_message)
       {
          common::unittest::Trace trace;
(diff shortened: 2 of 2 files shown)


middleware/common/include/common/communication/device.h
```diff
@@ -259,11 +259,8 @@ namespace casual
 
             auto complete = algorithm::find_if( m_cache, [&]( auto& complete){ return complete.correlation() == correlation;});
 
-            if( complete)
+            if( complete && complete->complete())
```
lazan, line 262: Ahh, nice! A lot better then what I thought we had to do! God work man!

middleware/common/unittest/source/communication/test_ipc.cpp
```diff
@@ -598,5 +597,45 @@ namespace casual
...
+            transport.message.header.offset = std::distance( std::begin( complete.payload), part_begin);
+
+            native::blocking::send( destination.socket(), destination.address(), transport);
+
+            return part_end;
+         };
+
+         // The sender sends the first half of the complete message
+         auto part_begin = send_single_transport( complete, std::begin( complete.payload));
+
```
lazan, line 629: Do a `device.flush()` just in case we run on a system that only have "space" for one transport at a time.

middleware/common/unittest/source/communication/test_ipc.cpp
```diff
@@ -598,5 +597,45 @@ namespace casual
...
+         // The receiver discards the message by correlation
+         device.discard( correlation);
+
+         // Send the second half
+         send_single_transport( complete, part_begin);
+
+         // There should be no message for us to receive since the receiver has discarded it
+         unittest::Message receive_message;
+         EXPECT_NO_THROW({ EXPECT_FALSE( device::non::blocking::receive( device, receive_message, correlation)); });
```
lazan: It feels kinda cramped with two gtest "predicate" in one line.  `EXPECT_NO_THROW` as a "scope"?

middleware/common/unittest/source/communication/test_ipc.cpp
```diff
@@ -598,5 +597,45 @@ namespace casual
...
+
+            return part_end;
+         };
+
+         // The sender sends the first half of the complete message
+         auto part_begin = send_single_transport( complete, std::begin( complete.payload));
+
+         // The receiver discards the message by correlation
+         device.discard( correlation);
+
```
lazan, line 632: Or `device.flush()`. here. At least before we send another one.

lazan (commented):
An easy but brilliant fix. I have a few pointer on the the test-case though.

middleware/common/unittest/source/communication/test_ipc.cpp
AlbsAgren: Yeah, I wasn't sure about the NO_THROW, but it felt weird not having it since thats what failed pre-fix. But having it as a scope might be tidier.

lazan (approved):
Very good work.

PR #322, opened 2024-02-26
Patch/1.6/http inbound corrid

AlbsAgren:


middleware/service/include/service/manager/admin/model.h
```diff
@@ -201,6 +201,21 @@ namespace casual
...
+         common::strong::correlation::id correlation;
+
+         CASUAL_CONST_CORRECT_SERIALIZE(
+            CASUAL_SERIALIZE( service);
+            CASUAL_SERIALIZE( caller);
+            CASUAL_SERIALIZE( callee);
+            CASUAL_SERIALIZE( correlation);
+         )
+      };
+
```
lazan, line 218: I think this will be usefull.

middleware/service/source/manager/state.cpp
```diff
@@ -123,6 +123,14 @@ namespace casual
                   m_service = replacement;
             }
 
+            std::optional< std::string> Sequential::reserved_service() const
+            {
+               if( m_service != nullptr)
+                  return m_service->information.logical_name();
+
```
lazan, line 130: nitpick, I would rather see `if( ! m_service)`. Why? more generic. If we change `m_service` to be, say an optional, it'll still work.

lazan: ` != nullptr` is perfectly fine though.

middleware/service/source/manager/transform.cpp
```diff
@@ -120,6 +120,26 @@ namespace casual
...
             }
 
+            auto reservation()
+            {
+               return []( auto& pair)
+               {
+                  auto& instance = std::get< 1>( pair);
+
+                  manager::admin::model::Reservation result;
+                  result.service = instance.reserved_service().value_or( result.service);
```
lazan, line 130: Is this a good paradigm? I had to look an extra time to understand that it is just _value or default_. 
Hmm, it gets ugly to do it any other way...
We should have an _algorithm_ for this. like `coalesce` but returns `T` default initialised if all _optional like_ is empty.

lazan (approved):

middleware/service/source/manager/state.cpp
AlbsAgren: No I agree with you, I'll update

middleware/service/source/manager/transform.cpp
AlbsAgren: This seems to be the pattern we've used thus far (see for example configuration/model/transform.cpp), but yeah, maybe it should be an algorithm.

lazan (approved):

PR #325, opened 2024-02-27
domain: improve error messages

### LINKED ISSUE #315: Improve error messages when casual is unable to create domain lock file
When the casual domain is started and casual is unable to create or write to the domain lock file, the error messages doesn't really explain the error. Sometimes they're even misleading, such as "failed to remove file", when the actual problem was encountered during creation.

partickl:
Add more descriptive error messages and raise error if domain lock file cannot be created

Resolves #315

### DIFF
--- a/middleware/common/source/domain.cpp
+++ b/middleware/common/source/domain.cpp
@@ -12,6 +12,7 @@
 
 #include "common/code/raise.h"
 #include "common/code/casual.h"
+#include "common/code/system.h"
 
 #include "common/serialize/json.h"
 
@@ -129,10 +130,16 @@ namespace casual
 
             file::scoped::Path temp_file{ file::name::unique( path.string(), ".tmp")};
             {
-               std::ofstream output( temp_file);   
+               std::ofstream output( temp_file);
+               if ( ! output)
+                  code::raise::error( code::casual::invalid_file, "cannot create domain lock file: ", temp_file.string(), ", errno: ", code::system::last::error());
+
                auto archive = serialize::json::writer();
                archive << model;
                archive.consume( output);
+
+               if ( ! output)
+                  code::raise::error( code::casual::invalid_file, "cannot write to domain lock file: ", temp_file.string(), ", errno: ", code::system::last::error());
             }
 
             log::line( log::debug, "domain singleton model: ", model);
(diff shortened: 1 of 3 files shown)


middleware/common/source/file.cpp
```diff
@@ -133,7 +133,11 @@ namespace casual
 
          void rename( const std::filesystem::path& source, const std::filesystem::path& target)
          {
-            std::filesystem::rename( source, target);
+            std::error_code ec;
+            std::filesystem::rename( source, target, ec);
+            if (ec)
+               code::raise::error( code::casual::invalid_path, "failed to rename file, source: ", source, " -> target: ", target, ", error message: ", ec.message());
+
```
lazan, line 140: Nice.

middleware/common/source/domain.cpp
```diff
@@ -129,10 +130,20 @@ namespace casual
...
-               std::ofstream output( temp_file);   
+               std::ofstream output( temp_file);
                auto archive = serialize::json::writer();
                archive << model;
                archive.consume( output);
+
+               if ( ! output)
+               {
+                  auto ec =  std::make_error_code( code::system::last::error());
+
```
lazan: I don't think it's conformant to assume _failed std::ofstream_ sets posix errc. Or is it?
If not, just skip the "error message" in the raised error.

middleware/common/source/domain.cpp
```diff
@@ -129,10 +130,20 @@ namespace casual
...
+                  if ( std::error_code temp_ec; ! std::filesystem::exists( temp_file, temp_ec))
+                     temp_file.release();
```
lazan: I don't understand this. If the file does not exist, we release the _temp_file_. Why? Do we get an error in `scoped::Path` dtor otherwise?

partickl: yes. if the path is non-empty it will try to remove the (non-existing) file. I'll move this check to the destructor for `scoped::Path`  instead

middleware/common/source/domain.cpp
partickl: seems to only be guaranteed when actually running on a posix compliant system.

lazan (approved):
Nice.

PR #329, opened 2024-03-04
Feature/1.7/branch on inbound

lazan:
This is a huge PR. I'm not sure how we're going to review this...


middleware/common/include/casual/task/concurrent.h
lazan, line 1: Hmm, this should not be in this commit...

middleware/common/include/common/log/line.h
```diff
@@ -0,0 +1,46 @@
+//!
```
AlbsAgren, line 1: Neat!

middleware/common/include/common/message/transaction.h
```diff
@@ -19,388 +20,399 @@
...
+            } // involved
+         } // external
+      } // resource
 
-            template<>
-            struct type_traits< transaction::resource::configuration::Request> : detail::type< transaction::resource::configuration::Reply> {};
+      namespace coordinate::inbound
+      {
+         using base_request = message::basic_request< message::Type::transaction_coordinate_inbound_request>;
+         struct Request : base_request
```
AlbsAgren, line 373: This is to request a branch right? Isnt something like inbound::branch::Request like in the commit message more descriptive?

middleware/common/include/common/strong/span.h
```diff
@@ -0,0 +1,38 @@
...
+{
+   namespace common::strong
+   {
+      template< typename T, typename Tag>
+      struct Span : std::span< T>
+      {
+         using base_type = std::span< T>;
+         using base_type::base_type;
+
+         //template< std::contiguous_iterator Iter>
```
AlbsAgren: Can this be removed?

thirdparty/database/include/sql/database.h
```diff
@@ -149,6 +149,12 @@ namespace sql
          return code::make( sqlite3_bind_blob( statement, index, value.data(), value.size(), SQLITE_STATIC));
       }
 
+      template< casual::concepts::binary::like Range>
```
AlbsAgren, line 128: Dont know if it matters, but I'm guessing this change is supposed to be in [sql: generalized parameter_bind overload for binary like range](https://github.com/casualcore/casual/pull/329/commits/350e7320801ed017a9d3a54c7cf5f1e7ab06455c)?

middleware/common/include/common/strong/id.h
```diff
@@ -55,9 +50,32 @@ namespace casual
          } // descriptor
       } // file
 
-      namespace socket
+      namespace ipc
```
AlbsAgren, line 53: This feels very helpful!

middleware/domain/source/discovery/state.cpp
```diff
@@ -32,7 +32,7 @@ namespace casual
          void Providers::registration( const message::discovery::api::provider::registration::Request& message)
          {
             // we only add 'new' processes
```
AlbsAgren, line 34: Very nitpicky, but now we're adding devices right?

middleware/transaction/source/manager/handle.cpp
```diff
@@ -400,35 +489,46 @@ namespace casual
...
+                           common::log::error( common::code::casual::invalid_semantics, " prepare callback - failed to find transaction 'context' for: ", origin, " - action: ignore");
+                           common::log::verbose::error( common::code::casual::invalid_semantics, "replies: ", replies, ", outcome: ", outcome);
 
                            return;
                         }
 
+                        // some sanity checks
+                        {
+                           if( transaction->stage != state::transaction::Stage::prepare)
+                              common::log::error( common::code::casual::invalid_semantics, "transaction is not in prepare phase - actual: ", transaction->stage, " - action: ignore");
```
AlbsAgren, line 496: Should there be a return here seeing as the the log message says `action - ignore`?

middleware/service/unittest/source/manager/test_state.cpp
```diff
@@ -249,14 +249,14 @@ namespace casual
...
+            auto process = service->reserve_concurrent( {});
             EXPECT_TRUE( process.pid == common::strong::process::id{ 103});
             
             // expect only 103 to be in the prioritized range
-            EXPECT_TRUE( process == service->reserve_concurrent());
+            EXPECT_TRUE( process == service->reserve_concurrent( {}));
 
             // we remove the instance
-            service->remove( )
+            //service->remove( )
```
AlbsAgren: 🧹

middleware/queue/source/manager/state.cpp
```diff
@@ -90,15 +90,37 @@ namespace casual
          }
       }
 
+      namespace local
+      {
+         namespace
+         {
+            void remove_queues( State& state, common::process::compare_equal_to_handle auto id)
```
AlbsAgren, line 97: You can now 'constrain' auto? Fancy

middleware/queue/source/forward/handle.cpp
```diff
@@ -926,25 +931,93 @@ namespace casual
...
+
+                           if( forward.target != handle)
+                              return;
+
+                           forward.target.invalidate();
+
+                           // if we've got pending enqueues to lost target we need to fake a reply
+                           // to keep the "state-machine" going
+                           for( auto& pending : algorithm::filter( state.pending.enqueues, is_forward( forward.id)))
+                              detail::push::faked::reply( pending);
```
AlbsAgren, line 988: This feels a bit unfortunate, but I can't think of a better way to do it

middleware/gateway/include/gateway/group/inbound/tcp.h
```diff
@@ -21,6 +21,7 @@ namespace casual
          return group::tcp::send( state, &handle::connection::lost, descriptor, std::forward< M>( message));
       }
 
+/*
```
AlbsAgren: Can this be removed?

middleware/gateway/unittest/source/group/outbound/test_state.cpp
```diff
@@ -15,11 +15,12 @@ namespace casual
 
    namespace gateway::group::outbound
    {
+      /*
```
AlbsAgren: Can this be removed?

AlbsAgren (commented):
Feels like a major improvement overall!

thirdparty/database/include/sql/database.h
lazan: Yes, probably :)

middleware/service/unittest/source/manager/test_state.cpp
lazan: Check.

middleware/queue/source/manager/state.cpp
lazan: Abbreviated function template. I think it's really nice. A lot less verbose, and it gets pretty self explanatory with a concept.

middleware/queue/source/forward/handle.cpp
lazan: I concur...

middleware/gateway/unittest/source/group/outbound/test_state.cpp
lazan: Check.

middleware/transaction/source/manager/handle.cpp
lazan: We might terminate. Something is wrong with our own code...

middleware/gateway/include/gateway/group/connection.h
lazan, line 1: Remove, or change external::tcp... to this...

middleware/gateway/include/gateway/group/inbound/state.h
```diff
@@ -339,8 +212,8 @@ namespace casual
             CASUAL_SERIALIZE( directive);
             CASUAL_SERIALIZE( external);
             CASUAL_SERIALIZE( pending);
-            CASUAL_SERIALIZE( correlations);
-            CASUAL_SERIALIZE( conversations);
+            //CASUAL_SERIALIZE( correlations);
+            //CASUAL_SERIALIZE( conversations);
```
lazan: Remove

middleware/gateway/include/gateway/group/inbound/tcp.h
lazan: Remove!

middleware/gateway/source/group/connection.cpp
lazan, line 1: Remove, or move tcp to this.

middleware/gateway/source/group/inbound/task/create.cpp
```diff
@@ -215,82 +213,161 @@ namespace casual
...
+                     // If the call failed, we've "sent" the reply, 
                      if( ! shared->origin_trid || shared->origin_trid != shared->message.trid)
-                        local::send::service_request( state, *shared->lookup, shared->message);
+                        local::send::service_request( state, descriptor, *shared->lookup, shared->message);
                      
                      return casual::task::concurrent::unit::Dispatch::pending;
 
                   },
-                  [ &state, shared]( common::message::transaction::coordinate::inbound::Reply& reply) mutable
+                  [ &state, shared]( common::message::transaction::coordinate::inbound::Reply& reply, strong::socket::id descriptor) mutable
```
lazan: Change name to transaction::branch::Request... ?

middleware/gateway/source/group/outbound/handle.cpp
```diff
@@ -40,133 +40,77 @@ namespace casual
 
             namespace internal
             {
+               namespace precondition
+               {
+                  bool reply( State& state, strong::socket::id descriptor, auto& message, auto&& create_reply)
+                  {
```
lazan, line 47: Better name?

middleware/gateway/source/group/outbound/handle.cpp
```diff
@@ -717,159 +555,118 @@ namespace casual
...
+                              state.pending.transactions.remove( message.trid, descriptor);
 
-                              state.coordinate.transaction.prepare( message);
-                           };
-                        }
-                           
-                     } // prepare
+                           auto destination = state.reply_destination.extract( message.correlation);
+                           state.multiplex.send( destination.ipc, message);
+                        };
```
lazan, line 594: Why not use the transaction dissasociate::transaction event?

middleware/gateway/include/gateway/group/inbound/state.h
OlofGullnas: There is another case of a commented CASUAL_SERIALIZE a bit up in the file (around line 174)

middleware/queue/source/forward/handle.cpp
```diff
@@ -926,25 +931,93 @@ namespace casual
...
-                        message.correlation = pending.correlation;
-                        return message;
+                        // Note: We rely on others to send us "error replies" for the following 
+                        //  * pending.service.calls : service-manager, gateway::outbound
+                        //  * pending.transaction.(commit|rollback) : transaction-manager
+
+                        log::line( verbose::log, "state.pending: ", state.pending);
+
+                        static constexpr auto is_forward = []( auto id) { return [ id]( auto& forward){ return forward == id;};};
+
```
OlofGullnas, line 965: I built this branch with g++-13 (with my changes for gcc-13 feature/1.7/main rebased to this branch). 
I got a warning/error:
`compile: /export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp
error:
/export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp: In instantiation of ‘void casual::queue::forward::local::{anonymous}::handle::dead::detail::dead_event(casual::queue::forward::State&, auto:153) [with auto:153 = casual::common::strong::Type<int, casual::common::strong::detail::integral::policy<int, casual::common::strong::process::tag, -1> >]’:
/export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp:1029:43:   required from here
/export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp:964:47: error: variable ‘is_forward’ set but not used [-Werror=unused-but-set-variable]
  964 |                         static constexpr auto is_forward = []( auto id) { return [ id]( auto& forward){ return forward == id;};};
      |                                               ^~~~~~~~~~
/export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp: In instantiation of ‘void casual::queue::forward::local::{anonymous}::handle::dead::detail::dead_event(casual::queue::forward::State&, auto:153) [with auto:153 = casual::common::strong::Type<casual::common::Uuid, casual::common::strong::ipc::tag>]’:
/export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp:1041:43:   required from here
/export/home/guo/c_1.7/casual-1.7/middleware/queue/source/forward/handle.cpp:964:47: error: variable ‘is_forward’ set but not used [-Werror=unused-but-set-variable]
cc1plus: all warnings being treated as errors

aborting due to errors`
I can't say if this is a legitimate warning or not. I looked arounf a little in bugzilla for gcc/g++ and there seemed to be som issues having to do with  constexpr and parameters and lambda captures in the gcc-9 to gcc-11 timeframe that has been corrected.
I removed static and added "explicit" capture of "is_forward" to the invalidate_queue_forward and invalidate_service_forward lambdas. Then code builds without warings. And unit tests run.

middleware/common/include/common/message/transaction.h
lazan: I concur.

middleware/queue/source/forward/handle.cpp
lazan: You should be able to create a constexpr-static non capturing lambda and it behaves just as a free function. Maybe it needs to be `inline` ?

middleware/common/include/common/network/byteorder.h
```diff
@@ -29,18 +30,12 @@ namespace casual
...
-               {
-                  static std::uint8_t encode( bool value) noexcept;
-                  static bool decode( std::uint8_t value) noexcept;
-               };
-
-               template<>
-               struct transcode<char>
+               // just passthrough 
+               template< typename T> requires concepts::any_of< T, bool, char, signed char, unsigned char>
+               struct transcode< T>
```
OlofGullnas, line 35: Should this (for completeness since c++20) include char8_t in the "any_of" list of types?

In C char8_t is a "typedef char8_t unsined char", but I think it is a distinct type in C++ 20.
 I don't think char8_t is used in casual so this comment is somewhat academical...

middleware/queue/source/forward/handle.cpp
OlofGullnas: I (!) think it is a compiler bug. Turning off the warning is a possibility, but it may be useful and correct in other cases. As far as I know this is the only place the warning is produced with casual.

For some reason the compiler misses that the variable "static constexpr is_forward" is used by  the lambdas invalidate_queue_forward and invalidate_service_forward. Some of the problems they fixed in earlier g++ versions seemed to have to do with the compiler missing to "mark" the use of some variables on some paths through the compiler. There might be one more for the case when a static constexpr lambda is used by another lambda... I will try to reproduce in a small code example. Perhaps not easy. I do not know  how much context is needed...

thirdparty/database/include/sql/database.h
lazan: I'll let this slide...

middleware/queue/source/forward/handle.cpp
lazan: Try to use it without the constexpr. That should also work, but then the compiler needs to guard the initialization for concurrent threads.

middleware/common/include/common/network/byteorder.h
lazan: I concur. And char8_t should be used :) But I think we do this when (if) we use std::u8string "everywhere". It's up for debate... @KristianIvarsson and I have talked a bit about it. We should take a stance which "encoding" we're handling/supporting (now we don't "care", and the user needs to have the same encoding on all machines). If we take a stance, then utf8 it is -> std::u8string "everywhere".

middleware/queue/source/forward/handle.cpp
OlofGullnas: Already tried. Same compiler warning. Turning off warning or remove static and capture (by value is what I tried) is_forward variable seems to work.

middleware/gateway/source/group/outbound/handle.cpp
lazan: Future us will decide...

middleware/queue/source/forward/handle.cpp
lazan: High probability for a compiler bug. I'll say we lay low with g++13 for a while. At least until we chose to use it in our build pipeline.

AlbsAgren (approved):
Looks real good I think

lazan:
Merged manually

PR #330, opened 2024-03-06
gateway: fix markdown documentation for interdomain protocol

### LINKED ISSUE #319: gateway binary markdown documentation is not correct
In the generated documentation all integers are uint64, which is not what actually is sent over the wire. Some attributes are int32 and such, and is sent as uint32 over the wire. 
The documentation should show this, of course.

lazan:
Before: all int (32b scalars) where generated as uint64 in the documentation. Although, the actual bits that are sent over the wire are uint32.

I'm not sure if sending 32b where intentional. But this is what we've got since 1.4, and we can't (of course) change the protocol.

Now: The markdown documentation (generator) is producing the correct documentation for 32b types.

resolves #319


partickl (approved):

lazan:
Merged manually

PR #338, opened 2024-03-13
build: adapt pipeline to multiple runners

ahhud:
Has adapted github actions pipeline to handle multiple runners.

As previously stated, in order to get any beta or release build to work properly, this fix is needed. That is, merging to release or main-branch may fail until this fix is used.

### DIFF
--- a/.github/workflows/casual-build-beta.yaml
+++ b/.github/workflows/casual-build-beta.yaml
@@ -281,7 +281,7 @@ jobs:
           tag_prefix: ""
 
   publish-version:
-    runs-on: self-hosted
+    runs-on: [self-hosted, casual, publish]
     needs: [build-centos,build-suse,tag-repo]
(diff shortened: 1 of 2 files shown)


.github/workflows/casual-build-beta.yaml
```diff
@@ -281,7 +281,7 @@ jobs:
           tag_prefix: ""
 
   publish-version:
-    runs-on: self-hosted
+    runs-on: [self-hosted, casual, publish]
```
AlbsAgren, line 284: Is `publish` like a tag you've added to the original runner?

ahhud: Yes, Its any kind of label I have added to the runners "identification". And if you have an array of requested labels, all have to match the specific runner.

AlbsAgren (approved):
Neat!

PR #339, opened 2024-03-15
service: prevent instances about to be assassinated  being reserved

### LINKED ISSUE #305: service - manager issues reservations for instances that it has previously requested be assassinated
This causes an 'invalid-semantics' error when the SM receives the process::Exit event and realises another call is in progress.

AlbsAgren:
Before: if a call to a service with a 'lethal' timeout contract reached the timeout and then completed soon after, the SM would reserve this instance again even though it would soon die Now: instances are marked as condemned when a hit is put out on them, preventing them from being reserved

Resolves #305

### DIFF
--- a/middleware/service/include/service/manager/state.h
+++ b/middleware/service/include/service/manager/state.h
@@ -119,6 +119,10 @@ namespace casual
                //! @returns the name of the reserved service if a reservation exists
                std::optional< std::string> reserved_service() const;
 
+               //! marks this instance as soon to die and thus ineligible for service reservations
+               inline void condemn() { m_condemned = true;}
+               inline bool condemned() const { return m_condemned;}
+
                friend std::ostream& operator << ( std::ostream& out, State value);
 
                CASUAL_LOG_SERIALIZE(
@@ -126,6 +130,7 @@ namespace casual
                   CASUAL_SERIALIZE_NAME( m_service, "service");
                   CASUAL_SERIALIZE_NAME( m_caller, "caller");
                   CASUAL_SERIALIZE_NAME( m_services, "services");
+                  CASUAL_SERIALIZE_NAME( m_condemned, "condemned");
                )
 
             private:
@@ -133,6 +138,7 @@ namespace casual
                instance::Caller m_caller;
                // all associated services
                std::vector< state::Service*> m_services;
+               bool m_condemned = false;
             };
 
             struct Concurrent : base_instance
@@ -247,6 +253,8 @@ namespace casual
 
                   inline const common::process::Handle& process() const { return get().process;}
                   inline auto state() const { return get().state();}
+
+                  inline bool condemned() const { return get().condemned();}
 
                   inline friend bool operator == ( const Sequential& lhs, common::strong::process::id rhs) { return lhs.process().pid == rhs;}
 
--- a/middleware/service/source/manager/handle.cpp
+++ b/middleware/service/source/manager/handle.cpp
@@ -18,6 +18,7 @@
 #include "common/algorithm.h"
 #include "common/algorithm/is.h"
 #include "common/algorithm/coalesce.h"
+#include "common/algorithm/compare.h"
 #include "common/process.h"
 #include "common/message/dispatch.h"
 #include "common/message/dispatch/handle.h"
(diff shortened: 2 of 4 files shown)


middleware/service/source/manager/handle.cpp
```diff
@@ -931,14 +945,14 @@ namespace casual
...
                   log::line( verbose::log, "state.timeout_instances: ", state.timeout_instances);
-
-                  if( ! ack.metric.trid)
-                     return;
                   
                   if( auto found = algorithm::find( state.timeout_instances, ack.metric.process.pid))
                   {
                      state.timeout_instances.erase( std::begin( found));
 
+                     if( ! ack.metric.trid)
```
AlbsAgren, line 953: Someone should check my work on this, but it didn't seem to make sense that we add instances to `timeout_instances` regardless of whether they're in a transaction, but only remove them if they are.

middleware/service/include/service/manager/state.h
```diff
@@ -119,20 +119,26 @@ namespace casual
                //! @returns the name of the reserved service if a reservation exists
                std::optional< std::string> reserved_service() const;
 
+               //! marks this instance as soon to die and thus ineligible for service reservations
+               inline void condemn() { m_condemned = true;}
```
ahhud, line 123: Nice names! :-)

ahhud (approved):

middleware/service/include/service/manager/state.h
lazan: Very nice!

middleware/service/source/manager/state.cpp
```diff
@@ -299,7 +299,12 @@ namespace casual
             const common::process::Handle& caller, 
             const strong::correlation::id& correlation)
          {
-            if( auto found = algorithm::find_if( instances.sequential, []( auto& i){ return i.idle();}))
+            auto reservable = [ &]( auto& instance)
+            {
+               return instance.idle() && ! instance.condemned();
+            };
+
+            if( auto found = algorithm::find_if( instances.sequential, reservable))
```
lazan, line 307: Easy to understand.

middleware/service/source/manager/handle.cpp
lazan: This must be better.

middleware/service/source/manager/handle.cpp
```diff
@@ -150,6 +151,15 @@ namespace casual
...
 
+      namespace detail::timeout::contract
+      {
+         using contract_type = common::service::execution::timeout::contract::Type;
+         bool lethal( contract_type contract)
+         {
+            return algorithm::compare::any( contract, contract_type::kill, contract_type::terminate);
+         }
+      }
+
```
lazan, line 162: Precise nomenclature!

middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1143,5 +1143,140 @@ namespace casual
...
+         try
+         {
+            common::process::sleep( std::chrono::seconds{ 10});
+         }
+         catch( ...)
+         {
+            auto error = common::exception::capture();
+            EXPECT_TRUE( error.code() == common::code::signal::terminate);
+         }
+
```
lazan, line 1183: Hmm... This got me thinking... `terminate` is by definition not lethal. 
Are we thinking about this problem the wrong way?... 

This patch assumes that if a _lethal_ assassination is sent, we'll wait for the process::Exit event. This might not happen... 

I've got second thoughts about this... If this patch does not fix the problem 100% then we shouldn't apply it. As I see it the patch might make it worse, at least for `terminate`.

Your thoughts @AlbsAgren?

AlbsAgren: Since this issue is a pretty extreme edge-case to begin with I'm inclined to agree. A bigger rethinking of how we handle assassinations in 1.7 might be a better way to go...

AlbsAgren: I guess the alternative would be to 'rehabilitate' instances that get SIGTERM and then ACK. Though not super elegant...

AlbsAgren:
Closing this as we rethink how to solve the problem.

PR #340, opened 2024-03-16
common: improve flag abstraction to use only pure enums

lazan:
A lot cleaner and quite "clever" (subjective) reduction of the abstraction.

Using _declare a function with a specific name_ to opt in if a `enum` should be treated as a flag-like. Quite terse, and one can opt in directly where the `enum` is defined. Using "classic" type-specialization to opt in, one (often) need to opt in "far" from the type. I'm not sure which is better... time will tell.

### DIFF
--- a/middleware/common/include/common/flag.h
+++ b/middleware/common/include/common/flag.h
@@ -6,176 +6,191 @@
 
 #pragma once
 
-#include "common/serialize/macro.h"
-#include "common/string.h"
-#include "common/traits.h"
+#include "common/algorithm.h"
+#include "common/stream/customization.h"
 
-#include "common/code/raise.h"
-#include "common/code/casual.h"
-
-#include <initializer_list>
-#include <type_traits>
+#include <bitset>
 
 namespace casual
 {
-   namespace common
+   namespace common::flag
    {
+      //! We're using strongly typed enums as flag abstraction. This makes it harder to
+      //! produce bugs.
+      //! 
+      //! To opt in to use an enum as flag declare:
+      //!
+      //! `consteval casual_enum_as_flag( <your enum type>);
+      //!
+      //! You can also opt in and use another flag as superset:
+      //!
+      //! `consteval <the superset type> casual_enum_as_flag_superset( <your enum type>);
+      //!
+      //! This will use the superset enum `description` function when printing your enum type.
+      //! You have to ensure that your enum type is a strict subset of the superset enum.
 
-      namespace has
+      namespace detail
       {
-         template< auto flags, typename T>
-         constexpr inline bool flag( T value)
+         template< typename T>
+         concept enum_as_flag = concepts::enumerator< T> && requires( T e)
          {
-            return ( value & flags) == flags;
-         }
-      } // has
+            casual_enum_as_flag( e);
+         };
 
-
-      template< typename E>
-      struct Flags
-      {
-         using enum_type = E;
-         using underlying_type = std::underlying_type_t< enum_type>;
-
-         static_assert( std::is_enum< enum_type>::value, "E has to be of enum type");
-         
-
-         //! just a helper to easier deduce the enum-type
-         constexpr static auto type() noexcept { return enum_type{};}         
-
-         constexpr Flags() = default;
-
-         template< typename... Enums>
-         constexpr Flags( enum_type e, Enums... enums) noexcept : Flags( bitmask( e, enums...)) {}
- 
-         constexpr explicit Flags( underlying_type flags) noexcept : m_flags( flags) {}
-
-         constexpr Flags convert( underlying_type flags) const
+         template< typename T>
+         concept has_description = enum_as_flag< T> && requires( T e)
          {
-            if( flags & ~underlying())
-               code::raise::error( code::casual::invalid_argument, "flags: ", flags, " limit: ", *this);
+            { description( e)} -> std::same_as< std::string_view>;
+         }; 
 
-            return Flags{ flags};
+         template< typename T>
+         concept enum_as_flag_superset = concepts::enumerator< T> && requires( T e)
+         {
+            { casual_enum_as_flag_superset( e)} -> has_description;
+         };
+
+         template< typename T>
+         concept can_use_description = has_description< T> || enum_as_flag_superset< T>; 
+
+
+         template< can_use_description T>
+         constexpr auto description_type( T flag)
+         {
+            if constexpr( enum_as_flag_superset< T>)
+               return static_cast< decltype( casual_enum_as_flag_superset( T{}))>( flag);
+            else
+               return flag;
          }
 
-         template< typename E2>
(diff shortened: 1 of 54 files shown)


middleware/common/source/communication/tcp.cpp
```diff
@@ -85,11 +85,14 @@ namespace casual
                   canonical = AI_CANONNAME,
                };
 
+               //! incicate that the enum is used as a flag (g++11 force to deinfe it)
```
ahhud: Spelling: _indicate_, _define_

middleware/common/source/communication/tcp.cpp
```diff
@@ -85,11 +85,14 @@ namespace casual
...
+               [[maybe_unused]] consteval void casual_enum_as_flag( Flag) {};
+
```
ahhud, line 90: Just out of curiosity, why did we need to _define_ it here and not on the other places this construction is used?

lazan: Because of a bug in g++11 :) Which I think is triggered because the compiler can see that it is not used (unnamed ns). It is used compile-time though which `consteval` strongly indicate. This is why we need the `[[maybe_unused]]` attribute, which we should not need. Then g++11 say it is not _defined_, so I defined it... 
In short, I think this bug "appear" because of the _unnamed ns context_.

middleware/common/include/common/flag.h
```diff
@@ -6,176 +6,168 @@
...
-
-         constexpr Flags convert( underlying_type flags) const
-         {
-            if( flags & ~underlying())
-               code::raise::error( code::casual::invalid_argument, "flags: ", flags, " limit: ", *this);
+template< casual::common::flag::enum_flag_like T>
+constexpr auto operator ! ( T value)
+{
+   return std::to_underlying( value) == 0;
+}
```
OlofGullnas: Had to do some thinking about what this did/should do. My first thougt was incorrect, My first thought was that operator! for a flag like object should do som kind of flag inversion ("complement"), which is not reasonable.

I think this does the same thing as common>::flag::empty().

middleware/common/include/common/flag.h
```diff
@@ -6,176 +6,168 @@
...
+   namespace common::flag
+   {
+      namespace native
+      {
+         template< auto flag>
+         constexpr bool exists( auto flags)
          {
-            return std::bitset< std::numeric_limits< underlying_type>::digits>{ static_cast< unsigned long long>( m_flags)}.count();
+            return ( flags & flag) == flag;
          }
```
OlofGullnas: What is the intention/use of this? Casual builds  and the test suite executes without "common::flag::native::exists( auto flags)". Usage seems a bit odd to me. Seems to be a function template with a value template parameter?

middleware/common/include/common/flag.h
```diff
@@ -6,176 +6,168 @@
...
+                     return;
 
-         template< typename Enum>
-         static auto print( std::ostream& out, Flags< Enum> flags, traits::priority::tag< 1>) -> decltype( out << description( Enum{}))
-         {
-            // TODO maintainence: This can probably be done a lot better...
-            if( ! flags)
-               return out << "[]";
+                  // print the flag
+                  out << ' ' << description( static_cast< T>( unsigned_type( 1) << index));
```
OlofGullnas, line 172: Noted that a number of the defined "Flag enums" ha a "no_flags" value defined (==0). Is it possible to be smart and output that description, if there is one? I do not think this code does that. Probaly prints "[]". Perfectly acceptable, but...

middleware/common/include/common/flag.h
lazan: Yeah,  this might not be a good operator to overload. The thinking was be able to do `if( ! flag)`, but since you can't do `if( flag)` it's not symmetrical. We might be better of not to implement `! operator`

middleware/common/include/common/flag.h
lazan: This to be able to check "native" flags.  `flag::native::exist< TPNOTRAN>( some_ulong)`.  Replacement for `has::flag< FLAG>( value)` 
I apparently got rid of all the uses with untyped flag-checking, we might as well remove this.

middleware/common/include/common/flag.h
lazan: Ahh, yeah, thats a thought... I don't think it's possible though since we can't know at compile-time if a given value has a _string representation_. We can't know if  `description( flag)` will return `no_flags` or `<unknown>` (which is our convention for absent string representation). Well, we can test for `<unknown>` but that seems a bit dodgy, to rely on a convention.

lazan: But, we can do a compile-time opt in for _absent string representation_.
```cpp
enum struct Flag
{
   no_flags = 0,
   flag1 = 1,
   flag3 =  4,
};

consteval void casual_enum_as_flag( Flag);
consteval casual_enum_as_flag_has_empty( Flag);
```
Then we can chose to use `description( ...)` for empty flag compile-time for a give enum type.

But... A bit of a hassle for very little gain.

lazan:
@ahhud @OlofGullnas, do you have any more objections?

ahhud:
No. 🙂

lazan:
> No. 🙂

Well, why not approve?

middleware/common/include/common/message/service.h
```diff
@@ -508,7 +510,7 @@ namespace casual
                std::string parent;
 
                common::transaction::ID trid;
-               request::Flags flags;
+               request::Flag flags{};
```
OlofGullnas, line 513: Linguistic/naming question. Not specific to this piece of code:
Should the enum class be named like "Flag" or "Flags"? Singular or plural? Pre change plural seems to have been common (I have not checked everywher1e).  As multiple flags can be set plural is kind of nice. 
Seems as if Flags has been changed to Flag in many places...

middleware/common/source/communication/tcp.cpp
OlofGullnas: g++-13 does not need the [[maybe_unused]].
It gives the warning 
middleware/common/source/communication/tcp.cpp:89:31: error: ‘consteval void casual::common::communication::tcp::local::{anonymous}::socket::casual_enum_as_flag(Flag)’ declared ‘static’ but never defined [-Werror=unused-function]
   89 |                consteval void casual_enum_as_flag( Flag);
      |                               ^~~~~~~~~~~~~~~~~~~

if the function is not defined. If the function is defined it compiles withjout warning withoiut [[maybe_unused]]

middleware/common/include/common/message/service.h
lazan: Not changed per se, The enum was still named `Flag` before, but we needed the `Flags` as a flag type, to actually be able to act as "flags". Hence, the enum Flag (Flag::a, Flag::b and so on) but parameters would normally be named `flags`, but still the enum type `Flag`

PR #343, opened 2024-03-25
Feature/1.7/no env reset

### LINKED ISSUE #335: Do not set environment variables -> potentially leaks memory on linux
Resetting environment variables leaks memory on linux systems. We need to make sure that se don't do this.

First: We probably need it in unittests, which is fine since the memory leak is not going to be big. And we really need to test environment variable stuff for config and such. We could move/create unittest::environment::variable::set.
We could change the current variable::set to _variable::set_if_not_set_ 

As of now we only seam to set `CASUAL_DOMAIN_INFORMATION` and `CASUAL_FIELD_TABLE`. 
* `CASUAL_DOMAIN_INFORMATION` we might be ok with this, since it's only DM that sets this. We could use _variable::set_if_not_set_ for this.

`CASUAL_FIELD_TABLE` we don't have access to in `common`, well, we don't have access to `buffer::fielded`. We might be ok with just _variable::set_if_not_set_  for this also.

lazan:


lazan:
In some files a did collapse ns -> probably easiest to use _ignore whitespace_ when reviewing...

middleware/common/source/process.cpp
```diff
@@ -333,11 +362,56 @@ namespace casual
...
+
+            auto pid = local::spawn::invoke( path, attributes, arguments, environment, actions.get());
+
+            // close the unused write end
+            ::close( cout_pipe[ 1]);
+            ::close( cerr_pipe[ 1]);
+
+            Capture capture;
+
+            capture.standard.out = local::spawn::read_from_pipe( cout_pipe[ 0]);
```
AlbsAgren, line 405: Clever 👍

middleware/common/source/communication/instance.cpp
```diff
@@ -241,20 +241,21 @@ namespace casual
...
+                        }
                      }
 
                      try
                      {
                         auto process = instance::fetch::handle( identity.id, directive);
 
-                        if( process && ! identity.environment.empty())
-                           environment::variable::process::set( identity.environment, process);
+                        //if( process && ! identity.environment.empty())
```
AlbsAgren: Remove?

AlbsAgren (approved):
Looks good!

middleware/common/source/communication/instance.cpp
lazan: I'll keep it rather. I think I commented that when I was testing to see what the consequences where if we didn't set these environment variables for children (in essence).

PR #345, opened 2024-03-25
Feature/1.7/chrono simplification

KristianIvarsson:
Fixed some stuff and simplified some stuff

### DIFF
--- a/middleware/common/include/common/chronology.h
+++ b/middleware/common/include/common/chronology.h
@@ -92,18 +92,6 @@ namespace casual
          }
       };
 
-      namespace supersede
-      {
-         template<>
-         struct point< platform::time::point::type>
-         {
-            static void stream( std::ostream& out, const platform::time::point::type& time)
-            {
-               chronology::utc::offset( out, time);
-            }
-         };
-      } // supersede
-
    } // common::stream::customization::supersede
 
 } // casual
--- a/middleware/common/include/common/chronology.h
+++ b/middleware/common/include/common/chronology.h
@@ -92,14 +92,17 @@ namespace casual
          }
       };
 
-      template<>
-      struct point< platform::time::point::type>
+      namespace supersede
       {
-         static void stream( std::ostream& out, const platform::time::point::type& time)
+         template<>
+         struct point< platform::time::point::type>
          {
-            chronology::utc::offset( out, time);
-         }
-      };
+            static void stream( std::ostream& out, const platform::time::point::type& time)
+            {
+               chronology::utc::offset( out, time);
+            }
+         };
+      } // supersede
 
    } // common::stream::customization::supersede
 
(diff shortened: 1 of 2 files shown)


OlofGullnas:
When building this branch with g++-13 in my private environment I do not get the error that the github build produced. A #include <format> in chronology.h not finding the "format" header). Is a compiler with support for the c++20 chrono enhancements being useed in the github build? This support is in g++-13, but may have been in g++-12 also. It was not present in g++-11.

I get another error in queue/source/forward/handle.cpp line 978. g++-13 incorrectly warns about a is_forward being a "unused-but-set-variable". This is a g++-13 false warning/bug. I have reported the bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114450).
Adding [[maybe_unused]] supresses the warning.

I also get another failure in the unit tests:

[ RUN      ] cli_queue.attributes
/export/home/guo/c_1.7/casual-1.7/middleware/administration/unittest/source/cli/test_queue.cpp:79: Failure
Value of: std::regex_match( output, std::regex{ R"(.*available: 42.*\n)"})
  Actual: false
Expected: true
output: { type: cli_queue_message, correlation: fa8a93799da9409a94e580bb5d1131cc, body: { execution: 6b0d2245e4d943aeb4effc4eab653841, id: nil, attributes: { properties: , reply: , available: 1970-01-01 00:00:42.000000000}, payload: { type: X_OCTET/, data: @0xe96420, size: 7, capacity: 7}}}

[  FAILED  ] cli_queue.attributes (275 ms)

In the 1.6 version of casual the available attribute was formatted as "available: 42000000000" which matches the regexp.
In my g++-13 build "available: 1970-01-01 00:00:42.000000000" is probably the result of formatting a time point with the standard c++20 chrono formatting code. Note that the overriding chrono output is not used. It would have produced an output of "1970-01-01T01:00:42.000000+01:00." or something like that. I have not investigated why the chronology.h code is not used. I suspect chronology.h is not included in some context where the attribute is serialized/formatted.

In a private branch I have previously experimented with moving parts of the formatting code in chronology.h to stream.h. and to include chronology.h in stream.h. Then the casual specific formatting is used in this unit test. Note that stream.h has special formatting code for some other classes so it is not completely illogical to also have support for casual chronology formatting in this file, It is a bit unusual (but not illegal...)  to have pieces of namespace comon::stream::customization(::supesede) in chronology.h.

I also do not know what the intention is in the command used in the unittest:

   casual queue --attributes available 42s

I believe that the attribute "available" is a point in time. It is a bit odd to specify it as "42s". The result seems to be a point in time 42 seconds after the base of the Unix epoch (1970-01-01 00:00 GMT or UTC).

lazan:
@KristianIvarsson rebase on `origin/fetaure/1.7/main`, and `push --force`

KristianIvarsson:
> @KristianIvarsson rebase on `origin/fetaure/1.7/main`, and `push --force`

Done

KristianIvarsson:
The feature branch is killed and removed and a new one is created

PR #351, opened 2024-03-28
example: fix casual/example/work

### LINKED ISSUE #349: casual/example/work does nothing
When calling casual/example/work it is expected to take at least the specified amount of time, preferably doing som cpu intensive work for the duration of the call

Currently it will exit immediately, regardless of configured duration.

from logfile:
```
<domain>|e8e3e8275713454f91dfa8dbbe713fce|9278|139780810491840|casual-example-server||||information|example server started with arguments: [--work, 1s]
<domain>|e8e3e8275713454f91dfa8dbbe713fce|9278|139780810491840|casual-example-server||||information|sleep: 0ns
<domain>|e8e3e8275713454f91dfa8dbbe713fce|9278|139780810491840|casual-example-server||||information|work: 1000000000ns

```

make call and check statistics:

```
%> casual service --metric-reset
%> echo '{ "key" : "some", "value": "json"}' | casual buffer --compose ".json/" | casual call --service casual/example/work | casual buffer --extract
{ "key" : "some", "value": "json"}

%>casual service -ls | grep work
casual/example/work                                          example   D  join    90.000000            1  1  0.000046  0.000046  0.000046  0  0.000000   0   0  2024-03-19T11:32:34.156331+01:00
```

partickl:
Correct the busyloop so we actually wait for the intended duration
Add simple unittest

Resolves #349

### DIFF
--- a/middleware/example/server/source/example/work.cpp
+++ b/middleware/example/server/source/example/work.cpp
@@ -102,7 +102,7 @@ namespace casual
          {
             auto deadline = platform::time::clock::type::now() + local::global.work;
 
-            while( deadline < platform::time::clock::type::now())
+            while( deadline > platform::time::clock::type::now())
             {
                ; // no-op
             }
--- a/middleware/example/server/unittest/source/test_example.cpp
+++ b/middleware/example/server/unittest/source/test_example.cpp
@@ -1,0 +1,95 @@
+//!
+//! Copyright (c) 2024, The casual project
+//!
+//! This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+//!
+
+#include "common/unittest.h"
+
+#include "casual/xatmi.h"
+#include "domain/unittest/manager.h"
+
+namespace casual
+{
+   using namespace common;
+
+   namespace example
+   {
+      namespace local
+      {
+         namespace
+         {
+            namespace configuration
+            {
+               constexpr auto base = R"(
+system:
+   resources:
+      -  key: rm-mockup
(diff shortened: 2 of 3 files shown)


lazan:
Should this be done with 1.6? Hmm, it would be practical regarding test-cases I presume...

middleware/example/server/source/example/work.cpp
```diff
@@ -102,7 +102,7 @@ namespace casual
          {
             auto deadline = platform::time::clock::type::now() + local::global.work;
 
-            while( deadline < platform::time::clock::type::now())
+            while( deadline > platform::time::clock::type::now())
```
lazan, line 105: Good catch! 😄

middleware/example/server/unittest/source/test_example.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+      TEST( example_server, work_500ms__expect_at_least_500ms_elapsed_time)
+      {
+         common::unittest::Trace trace;
+         
+         auto domain = local::domain();
+
+         auto elapsed_time = local::timed_action(local::call_work);
+
+         EXPECT_TRUE( elapsed_time > std::chrono::milliseconds{ 500});
```
lazan: I would like to see a shorter duration. If we've got 10 of these it's 5s. 10ms should be sufficient?

lazan: ... that is, we don't want to end up running all unittests takes several minutes.

partickl: agreed, i'll shorten as much as possible.

lazan (approved):

PR #358, opened 2024-04-04
build: add new builder

ahhud:
Has added yet another builder

### DIFF
--- a/.github/workflows/casual-build-alpha.yaml
+++ b/.github/workflows/casual-build-alpha.yaml
@@ -108,7 +108,7 @@ jobs:
           path: './**/casual-webapp.zip'
   
   build-centos:
-    runs-on: self-hosted
+    runs-on: [self-hosted, builder]
     if: ${{ success() }}
     needs: [calculate-version,build-frontend]
     env:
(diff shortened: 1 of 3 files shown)


AlbsAgren (approved):

lazan (approved):

lazan:
I think there might be a problem with som sort of race condition if two action runs overlap, sort of. Do you wanna check that out before we merge this?

ahhud:
> I think there might be a problem with som sort of race condition if two action runs overlap, sort of. Do you wanna check that out before we merge this?

Please elaborate. Which actions do you see that can run overlap?

lazan:
I started patch/1.6/docum... and patch/1.6/main a few minutes later. The last one failed to get some "state" from github. It worked when I rerun it after the first was done.

ahhud:
Ok, so it had nothing to do with this commit but rather the general behaviour when you in your case still have three builders when your closet is still in play which this commit is mitigating?
Ok, I will have a look.

ahhud:
Looks more like the Larsbuilders sometimes have a problem with building frontend resulting i a missing artifact. 
As it is the frontend artifact isn't even used. So I think I will remove all trace of it. But it is a separate problem and not linked to this. So I merge this needed one. And create a different issue to do some cleanup and lifecycling of versions on helper actions.

PR #360, opened 2024-04-05
queue: add example server with enqueue and dequeue services

### LINKED ISSUE #348: Add enqueue and dequeue services to casual-example-server
Should be possible to enqueue to or dequeue from a casual queue. Actual queue should be given as a command line parameter

partickl:
Resolves #348

### DIFF
--- a/middleware/queue/source/example/main.cpp
+++ b/middleware/queue/source/example/main.cpp
@@ -1,0 +1,104 @@
+#include "common/argument.h"
+#include "common/communication/ipc.h"
+#include "common/communication/instance.h"
+#include "common/exception/guard.h"
+#include "common/message/dispatch.h"
+#include "common/message/dispatch/handle.h"
+#include "common/server/argument.h"
+#include "common/server/handle/call.h"
+#include "queue/api/queue.h"
+
+namespace casual
+{
+   namespace queue::example
+   {
+      namespace local
+      {
+         namespace
+         {
+            struct
+            {
+               std::string queue;
+            } settings;
+
+            namespace service
+            {
+               auto enqueue()
+               {
+                  return []( common::service::invoke::Parameter&& parameter)
+                  {
+                     queue::Message message;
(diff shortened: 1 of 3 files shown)


middleware/queue/source/example/main.cpp
```diff
@@ -0,0 +1,104 @@
...
+                  }
+               }};
+            }
+
+            void main( int argc, char **argv)
+            {
+               common::argument::Parse{
+                  R"(Example server)",
+                  common::argument::Option{ std::tie( settings.queue), {"--queue"}, "queue that casual/example/{enqueue,dequeue} should use"},
+               }(argc, argv);
```
lazan, line 78: I was thinking that one might want to have different queues for enqueue and dequeue, but the real "need" is probably to supply in the request what queue one wants to address.
We'll keep it simple for now :)

middleware/queue/source/example/main.cpp
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            common::server::Arguments services()
+            {
+               return {{
+                  {
+                     "casual/example/enqueue",
+                     service::enqueue(),
+                     common::service::transaction::Type::automatic,
+                     common::service::visibility::Type::discoverable,
```
lazan, line 60: Transactions will not used. Which is fine?

lazan: @partickl Do you expect transactions to "work"?

middleware/queue/source/example/main.cpp
partickl: yes I think we could do another take on that in the future, we might add a blocking dequeue, and possibility to set message attributes as well

middleware/queue/source/example/main.cpp
partickl: yes, I think it should work, since `common::server::handle::Call` should handle it.
Also made a simple test case for rolling back after enqueue

Issue #308, opened 2024-02-04
Update build pipeline and remove deprecated tools


ahhud:
Some of the tools used in github actions is marked as deprecated or needs to be upgraded.

```

calculate-version
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions-ecosystem/action-regex-match@v2. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

calculate-version
The following actions uses node12 which is deprecated and will be forced to run on node16: actions-ecosystem/action-regex-match@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default

calculate-version
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

calculate-version
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands

build-frontend
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

build-suse
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/download-artifact@v3, actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20

build-centos
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/download-artifact@v3, actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20


```

Issue #361, opened 2024-04-06
Remove frontend build and update versions on github actions dependecies.


ahhud:
Remove not used frontend build and update some of the actions dependencies to get rid of github action warnings

PR #362, opened 2024-04-06
build: remove frontend, update dependencies

ahhud:
Removed the unused frontend and updated versions on dependencies in github actions to loose some warnings.

Resolves: #308,#361


lazan (approved):
Looking good

PR #365, opened 2024-04-17
http: use route name in outbound when ack:ing service calls

### LINKED ISSUE #364: casual-http-outbound ignores routes
It ACK:s with the 'original' name, meaning there are no metrics for the route.

AlbsAgren:
This patch fixes an issue where the http-outbound would ack with the 'original' service name, rather than the route. This resulted in no metrics being collected for the route.

Resolves #364


middleware/http/unittest/source/outbound/test_request.cpp
```diff
@@ -0,0 +1,64 @@
...
+      {
+         const state::Node& node()
+         {
+            static const auto singleton = [](){
+               state::Node value;
+               value.discard_transaction = true;
+               value.url = "http://casual.laz.se/some/service";
+               value.headers = std::make_shared< const common::service::header::Fields>();
+               return value;
+            }();
```
lazan, line 32: I like the _lambda initialisation trick_

lazan (approved):
Nice!

ahhud (approved):

PR #368, opened 2024-04-19
http: improved error logging in http-outbound

### LINKED ISSUE #367: Improve logging in http-outbound when curl requests fail
Would be good to at least know which service and url is involved

AlbsAgren:
This patch adds additional information to the error that is logged when a curl request in the http-outbound fails

Resolves #367


ahhud (approved):

lazan (approved):
Nice!

PR #369, opened 2024-04-20
Feature/1.7/string simplification

KristianIvarsson:
Removed the (proprietary) type `string::utf8` in favor of the standard `std::u8string` and some minor refactoring in order to create less strings

The naming could possibly be better, but I couldn't come up with anything better a.t.m. (the libraries we use are a bit fiddly)


lazan (approved):
Looks good to me. Nice to get rid of our own temporary utf8 type.

PR #371, opened 2024-04-24
documentation: updated and removed invalid api-documentation

lazan:
Removed all incomplete/invalid api-documentation for our admin services. Documented how a user could _describe_ these services instead.


partickl (approved):
I think this looks good. Examples are clearer now as well

PR #376, opened 2024-05-10
Feature/1.7/boot strict

AlbsAgren:
Feels a big inelegant maybe, open to suggestions.

### DIFF
--- a/middleware/domain/source/manager/admin/cli.cpp
+++ b/middleware/domain/source/manager/admin/cli.cpp
@@ -155,7 +155,7 @@ namespace casual
                      std::vector< std::string> arguments;
 
                      if( ! pattern.empty())
-                     {                           
+                     {
                         arguments.emplace_back( "--configuration");
                         algorithm::container::append( pattern, arguments);
                      }
@@ -262,13 +262,15 @@ namespace casual
                      });
                   };
 
-                  auto format_restart = []( const P& e){
+                  auto format_restart = []( const P& e)
+                  {
                      if( e.restart) 
                         return "true";
                      return "false";
                   };
 
-                  auto format_restarts = []( const P& e){
+                  auto format_restarts = []( const P& e)
+                  {
                      return e.restarts;
                   };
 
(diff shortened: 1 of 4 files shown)


middleware/domain/source/manager/admin/cli.cpp
```diff
@@ -781,11 +788,13 @@ note: not all options has legend, use 'auto complete' to find out which legends
                         { "domain.manager.server.instances.running", string::compose( instances_count( state.servers, running))},
                         { "domain.manager.server.instances.scale.out", string::compose( instances_count( state.servers, scale_out))},
                         { "domain.manager.server.instances.scale.in", string::compose( instances_count( state.servers, scale_in))},
+                        { "domain.manager.server.instances.missing", string::compose( instances_difference( state.servers, all, running))},
```
lazan, line 791: Is `missing` the best name for this? Feels like it doesn't completely describes what it is. Hmm, maybe it's ok. 
Since we'll be living with this a long while :) It's probably worth a few brain cycles to discuss this more.

lazan: Apart from my reservation about the naming of the information stuff, it's looks good.

AlbsAgren: I took it from the term we use for forward instances, though maybe we don't expose it to the user there. Either way I'm open to better terminology

lazan: `exited` ? At least it's more telling than `missing` i think. 
Do you only count exited != `0` as _missing_? If so, we could use `failed`?

AlbsAgren: Right it's just calculating configured - running, but maybe we should take exit codes into consideration somehow?

lazan: I'm not sure... Let's go with `exited`? It's ok with `missing` if you prefer it.

lazan: Hmm, I think I'm missing the point :) exited is not correct :). `missing` is actually much better.  Sorry for wasting your time :)

lazan (approved):

lazan:
@AlbsAgren I was thinking. Do we have an _information_ for aggregated exited? I imagine that it would be useful?

PR #378, opened 2024-05-24
service: fix invisible route being created when updating service config

### LINKED ISSUE #370: editing the configuration of a service with --configuration-edit creates an empty route to that service
Reproduce:

Start a domain with the following config:
```yaml
domain:
  name: blah

  servers:
     instances: 1

  services:
    - name: casual/example/echo
      execution:
        timeout:
          duration: 1min
          contract: kill
```

Use `casual domain --configuration-edit` to edit the service (for instance, set the timeout to 2min).

Call `casual internal --state-dump` on the service manager, there will be an empty route:
```
routes: [[ casual/example/echo, []]]
```

AlbsAgren:
This patch fixes an issue where a route with no name would be created to a service whose configuration was changed during runtime. This would make it so that no additional instances of that service could be advertised.

Resolves #370

### DIFF
--- a/middleware/service/source/manager/admin/cli.cpp
+++ b/middleware/service/source/manager/admin/cli.cpp
@@ -182,7 +182,6 @@ namespace casual
                      }
 
                   } // service
-
                } // normalized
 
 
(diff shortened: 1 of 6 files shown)


middleware/service/source/manager/admin/cli.cpp
```diff
@@ -652,12 +660,24 @@ namespace casual
...
+                                 return format::route::Row{ service, route.target};
+                              });
 
-                        auto formatter = format::routes( state);
-                        formatter.print( std::cout, state.routes);
+                              return result;
+                           });
+                        };
+
+                        auto state = admin::api::state();
```
lazan: Make this `const`, since you hold the name/route by reference.

middleware/service/source/manager/configuration.cpp
```diff
@@ -164,7 +163,7 @@ namespace casual
 
                      if( auto found = algorithm::find( state.routes, configuration.name))
                         algorithm::append( add, found->second);
-                     else 
+                     else if( add)
```
lazan, line 166: This was the bug? Good catch.

lazan (approved):

middleware/service/source/manager/configuration.cpp
AlbsAgren: Aye :)

middleware/service/source/manager/admin/cli.cpp
AlbsAgren: Will do

PR #380, opened 2024-05-28
queue: enqueue/dequeue to muliple queues in example server

partickl:
Added functionality to configure the example server with multiple queues. Each queue will have a dedicated enqueue- and dequeue service.

### DIFF
--- a/middleware/queue/source/example/main.cpp
+++ b/middleware/queue/source/example/main.cpp
@@ -1,4 +1,5 @@
 #include "common/argument.h"
+#include "common/algorithm.h"
 #include "common/communication/ipc.h"
 #include "common/communication/instance.h"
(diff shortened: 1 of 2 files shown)


middleware/queue/source/example/main.cpp
```diff
@@ -50,41 +51,52 @@ namespace casual
...
+                     },
+                  };
+               };
+
+               auto services = common::algorithm::accumulate(
+                  common::algorithm::transform( queues, to_services),
+                  std::vector<common::server::Service>{},
+                  [](auto&& a, auto&& b) {
+                     return common::algorithm::append( b, a);
+                  });
```
lazan, line 81: Clever, and pretty easy to understand.

lazan (approved):
Looking good.

PR #382, opened 2024-06-10
Feature/1.7/interdomain protocol

lazan:
I tried to split this to several commits. Not entirely sure if each builds on their own. I think it's easier to review though.


middleware/gateway/source/manager/main.cpp
```diff
@@ -54,6 +54,10 @@ namespace casual
                return transform::state( casual::domain::configuration::fetch().gateway);
             }
 
+#if CASUAL_PROTOCOL_VERSION == 1002
```
AlbsAgren, line 57: What is happening here?

AlbsAgren (approved):

middleware/gateway/source/manager/main.cpp
lazan: Ahh, leftovers from playing with compiler defines.

PR #383, opened 2024-06-27
Feature/1.7/interdomain protocol

lazan:
A large PR... The binary-type commit is a large one and I'm not that concern about it since it's just refactoring and do what the compiler tells you.
The other commits needs a review at least.

### DIFF
--- a/middleware/http/source/outbound/transform.cpp
+++ b/middleware/http/source/outbound/transform.cpp
@@ -18,13 +18,14 @@ namespace casual
          {
             auto header( const std::vector< configuration::Header>& model)
             {
-               common::service::header::Fields headers;
+               auto transform_fields = []( auto& model)
+               {
+                  return common::algorithm::transform( model, []( auto& field){
+                     return common::service::header::Field{ field.name, field.value};
+                  });
+               };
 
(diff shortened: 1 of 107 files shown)


7AXEL:
Check out this project: https://github.com/7AXEL/Web2APK

middleware/http/source/outbound/transform.cpp
```diff
@@ -18,13 +18,14 @@ namespace casual
...
             auto header( const std::vector< configuration::Header>& model)
             {
-               common::service::header::Fields headers;
-
-               common::algorithm::transform( model, headers.container(), []( auto& h){
-                  return common::service::header::Field{ h.name, h.value};
-               });
-
-               return std::make_shared< const common::service::header::Fields>( std::move( headers));
+               auto transform_fields = []( auto model)
```
AlbsAgren: Very minor, but how come you aren't taking `model` by reference when the surrounding function does?

AlbsAgren (approved):

middleware/http/source/outbound/transform.cpp
lazan: Must be a mistake.

PR #386, opened 2024-07-15
Feature/1.7/queue enable disable

lazan:


middleware/gateway/source/group/outbound/handle.cpp
```diff
@@ -711,7 +711,13 @@ namespace casual
                            request.order = state.order;
                            request.queues.add = algorithm::transform( message.content.queues, []( auto& queue)
                            {
-                              return casual::queue::ipc::message::advertise::Queue{ queue.name, queue.retries};
+                              casual::queue::ipc::message::advertise::Queue result;
+                              result.name = queue.name;
```
KristianIvarsson, line 714: Use `const auto& queue` or `std::move() `

middleware/queue/source/manager/handle.cpp
```diff
@@ -472,7 +474,13 @@ namespace casual
                         for( auto& name : message.content.queues)
                         {
                            if( auto queue = get_queue( name))
-                              reply.content.queues.emplace_back( std::move( name), queue->order);
+                           {
+                              auto& value = reply.content.queues.emplace_back();
+                              value.name = name;
```
KristianIvarsson, line 426: Use `value.name = queue.name` and/or `std.move( name)` or preferable a transform-method

lazan:
merged manually

PR #391, opened 2024-08-04
Feature/1.7/gateway runtime configuration

lazan:


middleware/configuration/include/configuration/user.h
```diff
@@ -447,12 +447,13 @@ namespace casual
                {
                   std::optional< std::string> alias;
                   std::optional< std::string> note;
+                  std::optional< platform::size::type> order;
```
AlbsAgren, line 450: I take it we're going with order rather than using groups for priority then?

lazan: Exactly. After some thinking and discussing it's pretty clear that it would be too complicated for the user. The most basic priority would work with _grouping_, but if it's getting a little bit more complex setup, it would be too hard to understand. 
Then, how would we show the actual "order" to the user? 
The name `order` is up for debate though.

AlbsAgren: Spontaneously I'd think that `priority` makes it somewhat more clear what it does, but maybe theres a downside I havent considered.

AlbsAgren (approved):

middleware/configuration/include/configuration/user.h
lazan: Well, priority could indicate that there will be more traffic to high priority and less to low. Which is not the case, it's more _everything to the 'highest' order, and nothing to the rest_.
Or rather, equally distributed between all that has the same _'highest'_ order.

lazan: Also, we had `order` since before, no change in cli and such, which has to be a good thing.

lazan: We have discussed these:
* `order`
* `priority`  (nether @KristianIvarsson @ahhud and me thinks this is good)
* `rank`  (@ahhud proposal)

PR #392, opened 2024-08-11
Feature/1.7/caller includes pending time

lazan:


middleware/common/include/common/message/service.h
```diff
@@ -307,7 +320,6 @@ namespace casual
                   {
                      regular,
                      no_reply,
-                     no_busy_intermediate,
```
AlbsAgren, line 327: Feels nice to be rid of this, always found it a bit confusing.

middleware/common/include/common/service/lookup.h
```diff
@@ -18,95 +18,56 @@ namespace casual
...
-            ~Lookup();
+         } // non::blocking
+               
+      } // lookup
 
-            Lookup( Lookup&&) noexcept;
-            Lookup& operator = ( Lookup&&) noexcept;
+      struct Lookup
+      {
+         //Lookup() noexcept;
```
AlbsAgren: 🗑️

middleware/common/source/service/lookup.cpp
```diff
@@ -12,145 +12,111 @@ namespace casual
...
+         : Lookup( std::move( service), lookup::Context{}, std::move( deadline)) 
+      {}
 
-         void Lookup::update( Reply&& reply)
-         {
-            log::line( verbose::log, "reply: ", reply);
+      Lookup::~Lookup()
+      {
+         // we have to discard if we're still pending with the service manager
+         if( ! m_correlation)
```
AlbsAgren: If m_correlation is empty I dont think it makes sense to call local::discard since that function has a guard that returns if the correlation is empty? Should it be `if( m_correlation)`?

AlbsAgren (commented):
Doing it this way makes a lot of sense to me

middleware/common/source/service/lookup.cpp
lazan: Of course!

AlbsAgren (approved):

lazan:
Merged manually

PR #394, opened 2024-08-13
documentation: added operations tutorial

AlbsAgren:
This commit adds a tutorial that introduces basic
casual concepts from an operations perspective.

### DIFF
--- a/documentation/tutorials/operations/README.md
+++ b/documentation/tutorials/operations/README.md
@@ -1,0 +1,801 @@
+# casual Operations Tutorial
+
+## 1 Introduction
+
+This tutorial will cover the basics of casual from an operations perspective. We will go over how to configure a basic domain, how to add services and queues as well as setting up connections between different domains. More development-oriented topics like writing and building servers are outside the scope of this tutorial.
+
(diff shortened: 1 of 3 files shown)


documentation/tutorials/operations/README.md
```diff
@@ -0,0 +1,787 @@
...
+```
+
+#### 5.2 Forwards
+
+While Casual provides an API for executables to programmatically enqueue and dequeue messages, sometimes what is needed is an entity that simply continuously dequeues messages from a queue and either enqueues them onto another queue or uses their contents to call a service. To this end, Casual offers two flavours of *forward*: queues and services, that does just that.
+
+Let's return to our domain.yaml once more to add some forwards under queue:
+
+```yaml
+    forward:
```
lazan, line 390: We might want to at least show the "whole" model? 
```yaml
domain:
  # ...
  queue:
    # ...
    forward:
```
I imagine that this would be clearer.

lazan (commented):
I think this is really good! 
Impressive.
I do like you to spell casual as casual, and not Casual. I can live with it though 😃

documentation/tutorials/operations/README.md
AlbsAgren: Yeah, good idea!

lazan (approved):
Brilliant.

PR #395, opened 2024-08-15
Feature/1.7/queue group size

AlbsAgren:
I'm not quite sure about some decisions in this one, see comments etc., so I'm doing a draft PR to get some feedback.

### DIFF
--- a/middleware/common/include/common/quantity.h
+++ b/middleware/common/include/common/quantity.h
@@ -1,0 +1,67 @@
+//! 
+//! Copyright (c) 2024, The casual project
+//!
+//! This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+//!
+
+#pragma once
+
+#include "casual/platform.h"
+
+#include <string>
+
+namespace casual
+{
+   namespace common::quantity
+   {
+      namespace bytes
+      {
+         using bytes_type = platform::size::type;
+
+         namespace unit
+         {
+            constexpr bytes_type byte = 1;
+
+            constexpr bytes_type kilobyte = 1000;
+            constexpr bytes_type kibibyte = 1024;
+
+            constexpr bytes_type megabyte = 1'000'000;
+            constexpr bytes_type mebibyte = 1'048'576;
+
+            constexpr bytes_type gigabyte = 1'000'000'000;
+            constexpr bytes_type gibibyte = 1'073'741'824;
+
+            constexpr bytes_type terabyte = 1'000'000'000'000;
+            constexpr bytes_type tebibyte = 1'099'511'627'776;
+         } // unit
+
+         namespace symbol
+         {
+            constexpr auto byte = "B";
+
+            constexpr auto kilobyte = "kB";  // lower case intentional, see: IEC 80000-13
+            constexpr auto kibibyte = "KiB";
+
+            constexpr auto megabyte = "MB";
+            constexpr auto mebibyte = "MiB";
+
+            constexpr auto gigabyte ="GB";
+            constexpr auto gibibyte = "GiB";
+
+            constexpr auto terabyte = "TB";
+            constexpr auto tebibyte = "TiB";
+         } // symbol
+
+         namespace from
+         {
+            bytes_type string( std::string value);
+         } // from
+
+         namespace to
+         {
+            std::string string( bytes_type value);
+         } // to
+
+      } // bytes
+   } // common::quantity
+} // casual
--- a/middleware/common/source/quantity.cpp
+++ b/middleware/common/source/quantity.cpp
@@ -31,25 +31,34 @@ namespace casual
          {
             namespace
             {
+               struct map
+               {
+                  bytes_type unit;
+                  std::string_view symbol;
+                  
+                  bool operator == ( bytes_type rhs) const noexcept { return unit == rhs;}
+                  bool operator == ( std::string_view rhs) const noexcept { return symbol == rhs;}
+               };
+
+               constexpr std::array< map, 10> unit_mappings{{
+                  { unit::byte, ""},
+                  { unit::byte, symbol::byte},
+                  { unit::kilobyte, symbol::kilobyte},
+                  { unit::kibibyte, symbol::kibibyte},
+                  { unit::megabyte, symbol::megabyte},
+                  { unit::mebibyte, symbol::mebibyte},
+                  { unit::gigabyte, symbol::gigabyte},
+                  { unit::gibibyte, symbol::gibibyte},
+                  { unit::terabyte, symbol::terabyte},
+                  { unit::tebibyte, symbol::tebibyte}
+               }};
+
                namespace parse
                {
                   std::optional< bytes_type> symbol( std::string_view symbol)
                   {
-                     static const std::map< std::string_view, bytes_type> symbol_map{
-                        { "", unit::byte },
-                        { symbol::byte, unit::byte },
-                        { symbol::kilobyte, unit::kilobyte },
-                        { symbol::kibibyte, unit::kibibyte },
-                        { symbol::megabyte, unit::megabyte },
-                        { symbol::mebibyte, unit::mebibyte },
-                        { symbol::gigabyte, unit::gigabyte },
-                        { symbol::gibibyte, unit::gibibyte },
-                        { symbol::terabyte, unit::terabyte },
-                        { symbol::tebibyte, unit::tebibyte }
-                     };
-
-                     if( auto found = symbol_map.find( symbol); found != std::end( symbol_map))
-                        return found->second;
+                     if( auto found = algorithm::find( unit_mappings, symbol))
+                        return found->unit;
 
                      return std::nullopt;
                   }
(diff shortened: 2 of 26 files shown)


middleware/queue/include/queue/group/state.h
```diff
@@ -87,6 +87,21 @@ namespace casual
...
+
+            inline void add( platform::size::type size) { current += size;}
+            inline void subtract( platform::size::type size) { current -= size;}
+            inline platform::size::type available() const { return capacity - current;}
+
+            CASUAL_CONST_CORRECT_SERIALIZE(
+               CASUAL_SERIALIZE( current);
+               CASUAL_SERIALIZE( capacity);
+            )
+         } size;
```
lazan: Move it to `state::Size`, to keep the `State` type more clean.

middleware/queue/include/queue/manager/admin/model.h
```diff
@@ -27,6 +27,8 @@ namespace casual
          common::process::Handle process;
          std::string queuebase;
          std::string note;
+         platform::size::type size;
+         platform::size::type capacity;
```
lazan: Have a `Size` type?

middleware/queue/source/group/handle.cpp
```diff
@@ -209,20 +215,45 @@ namespace casual
...
+
                         try 
                         {
                            // Make sure we've got the quid.
                            message.queue = state.queuebase.id( message);
 
                            auto reply = state.queuebase.enqueue( message);
 
+                           state.size.add( message.message.payload.data.size());
+
```
lazan, line 265: Make a function that calculates the size of the queue-message? We then have one place where we can alter how we calculate the size (besides SQL statements).

middleware/queue/source/group/queuebase.cpp
```diff
@@ -598,7 +616,9 @@ namespace casual
          {
             m_statement.commit1.execute( id.range());
             const auto commit1_affected = m_connection.affected();
-            m_statement.commit2.execute( id.range());
+            // raises 'sql:another row available' if we dont fetch the result from RETURNING
+            // is this good or should we use another statement or should the function return std::tuple with affected and 'changed_size' instead?
+            sql::database::query::fetch( m_statement.commit2.query( id.range()), local::transform::size());
```
lazan: Hmm, this feels a bit so so...

middleware/queue/source/group/queuebase.cpp
```diff
@@ -627,27 +648,33 @@ namespace casual
...
 
          auto gtrid = common::transaction::id::range::global( id);
-         m_statement.rollback1.execute( gtrid);
+
+         auto sizes = sql::database::query::fetch( m_statement.rollback1.query( gtrid), local::transform::size());
+
          m_statement.rollback2.execute( gtrid);
          m_statement.rollback3.execute( gtrid);
+
+         return std::reduce( std::begin( sizes), std::end( sizes));
```
lazan, line 676: Should be in `algorithm` and take a range.

middleware/queue/source/manager/admin/cli.cpp
```diff
@@ -222,7 +222,9 @@ namespace casual
                      terminal::format::column( "alias", format_alias, terminal::color::yellow),
                      terminal::format::column( "pid", format_pid, terminal::color::white, terminal::format::Align::right),
                      terminal::format::column( "ipc", format_ipc, terminal::color::no_color, terminal::format::Align::right),
-                     terminal::format::column( "queuebase", format_queuebase, terminal::color::cyan)
+                     terminal::format::column( "queuebase", format_queuebase, terminal::color::cyan),
```
lazan, line 232: Get rid of  `ipc` while we're at it?

middleware/common/include/common/quantity.h
```diff
@@ -0,0 +1,67 @@
...
+            constexpr auto kibibyte = "KiB";
+
+            constexpr auto megabyte = "MB";
+            constexpr auto mebibyte = "MiB";
+
+            constexpr auto gigabyte ="GB";
+            constexpr auto gibibyte = "GiB";
+
+            constexpr auto terabyte = "TB";
+            constexpr auto tebibyte = "TiB";
```
lazan, line 52: Use `std::string_view`.

middleware/common/include/common/quantity.h
```diff
@@ -0,0 +1,67 @@
...
+         } // symbol
+
+         namespace from
+         {
+            bytes_type string( std::string value);
+         } // from
```
lazan, line 58: Should take `std::string_view` ?

middleware/common/source/quantity.cpp
```diff
@@ -0,0 +1,120 @@
...
+                        { unit::byte, symbol::byte},
+                        { unit::kilobyte, symbol::kilobyte},
+                        { unit::kibibyte, symbol::kibibyte},
+                        { unit::megabyte, symbol::megabyte},
+                        { unit::mebibyte, symbol::mebibyte},
+                        { unit::gigabyte, symbol::gigabyte},
+                        { unit::gibibyte, symbol::gibibyte},
+                        { unit::terabyte, symbol::terabyte},
+                        { unit::tebibyte, symbol::tebibyte}
+                     };
```
lazan: You could just have a _mapping type_ and use a constexpr `std::array`, and just linearly `algorithm::find`.
```cpp
struct map
{
   bytes_type unit;
   std::string_view symbol;
   
   inline bool operator == ( const map& lhs, bytes_type rhs) { return lhs.unit == rhs;)
   inline bool operator == ( const map& lhs, std::string_view rhs) { return lhs.symbol == rhs;)
};
```
And use the array for both to and from string.

middleware/configuration/include/configuration/user.h
```diff
@@ -658,6 +658,7 @@ namespace casual
                std::optional< std::string> alias;
                std::optional< std::string> queuebase;
                std::optional< std::string> note;
+               std::optional< std::string> capacity;
```
lazan: I think I'd rather have. 
```yaml
group:
   capacity:
      size: 42
      # we could add messages or something. Not likely maybe...
```
or
```yaml
group:
   limit:
      size: 42
      # or capacity: 42
```

or
```yaml
group:
   size:
      capacity: 42
```
or we keep it as is, and if we ever want to add other stuff we deprecate `capacity` and create som composite as above. Your call.

middleware/configuration/source/model/transform.cpp
```diff
@@ -870,6 +872,9 @@ namespace casual
                      result.alias = null_if_empty( value.alias);
                      result.queuebase = null_if_empty( value.queuebase);
 
+                     if( value.capacity)
+                        result.capacity = common::quantity::bytes::to::string( value.capacity);
+
```
lazan, line 882: I don't really like relying on implicit conversions to `bool` for _integers_. Should it be an `std::optional` instead? Then we should get that one could set a limit of `0B` and nothing could be enqueued to it. Might be a feature.

middleware/queue/include/queue/common/ipc/message.h
```diff
@@ -715,13 +715,18 @@ namespace casual
                std::vector< Queue> queues;
                std::vector< common::strong::queue::id> zombies;
 
+               platform::size::type size;
+               platform::size::type capacity;
+
```
lazan, line 735: Should be a composite?

lazan (commented):
Looking good. Nice unittests. 
The _consume row and throw it away thing_ feels a bit irritating though😄

middleware/common/include/common/quantity.h
AlbsAgren: I don't know, I need to mutate `value` to parse it, so I'd still end up making a copy somewhere I think

lazan: Ok.

middleware/queue/source/group/queuebase.cpp
AlbsAgren: Didn't really find a good way to merge this with 'regular' commit in the end, so I went for what I thought felt least bad.

middleware/queue/source/group/handle.cpp
```diff
@@ -209,20 +215,54 @@ namespace casual
...
                         log::line( verbose::log, "message: ", message);
 
+                        if( ! detail::has::sufficient::capacity( state, message))
+                        {
+                           log::line( log::category::error, " failed with enqueue request to queue: ", message.name, " - queue-group ", state.alias, " full");
+                           auto reply = common::message::reverse::type( message);
+                           reply.code = common::code::queue::no_queue;  // bespoke code for 'queuebase_full'?
+                           state.multiplex.send( message.process.ipc, reply);
+                           return;
+                        }
```
lazan, line 255: Nice.

lazan (approved):
Looking good!

lazan:
@AlbsAgren I was thinking, we might want to verify that the sqlite3-lib version is high enough during startup. 
`sql::database::version::library()` >= _the version returns was included_.
https://www.sqlite.org/c3ref/c_source_id.html

middleware/queue/source/group/main.cpp
```diff
@@ -60,6 +63,24 @@ namespace casual
 
                } // select
 
+               namespace sqlite::library::version
+               {
+                  constexpr int sqlite_version_number( int major, int minor, int patch)
+                  {
+                     return major * 1000000 + minor * 1000 + patch;
+                  }
```
lazan: Put this in _sql::database_ ? It feels like it belongs there. We might need to constrain for other versions in other managers later on.

AlbsAgren: Aye

thirdparty/database/include/sql/database.h
```diff
@@ -87,6 +87,14 @@ namespace sql
          {
             return sqlite3_libversion_number();
          }
+
+         namespace number
+         {
+            inline constexpr int calculate( int major, int minor, int patch)
+            {
+               return major * 1000000 + minor * 1000 + patch;
+            }
```
lazan, line 96: nice!

lazan (approved):

middleware/configuration/source/documentation/main.cpp
```diff
@@ -351,16 +351,27 @@ property                 | description                           | default
...
-alias : `string`       | the (unique) alias of the group.             | 
-[queuebase : `string`] | the path to the storage file.                | `domain.queue.default.directory/<group-name>.qb`
-[queues : `list`]      | defines all queues in this group, see below  |
+property                | description                                  | default
+------------------------|----------------------------------------------|------
+alias : `string`        | the (unique) alias of the group.             | 
+[queuebase : `string`]  | the path to the storage file.                | `domain.queue.default.directory/<group-name>.qb`
+[queues : `list`]       | defines all queues in this group, see below  |
+[capacity : `capacity`] | limits the storage capacity of the group     |
+
```
lazan, line 374: Did you forget to generate after you updated? 
Also, we should add the _capacity/size_ to the example model?

lazan: ... apparently I forgot to generate for the assassination contract update :)

PR #396, opened 2024-08-17
archive: simplification of implementation

KristianIvarsson:
Using new interface in 3rd party library in order to make xml archives a bit more simple and hopefully a bit more efficient


lazan (approved):
Looks good

PR #403, opened 2024-08-23
queue: casual queue --dequeue should take 0..* queue-ids

lazan:
This patch makes it possible to supply multiple queue-ids to `casual queue --dequeue`. This gives users more option when working with `casual queue` CLI


AlbsAgren (approved):
Good feature

PR #407, opened 2024-08-27
http: outbound always send metric on service call

### LINKED ISSUE #397: http outbound does not send metric/ack when http request fails immediately

AlbsAgren:
This commit ensures that the outbound always sends metrics to the service manager even when a call fails immediately due to violated preconditions.

Resolves #397

### DIFF
--- a/middleware/http/include/http/outbound/state.h
+++ b/middleware/http/include/http/outbound/state.h
@@ -205,6 +205,8 @@ namespace casual
          {
             void add( const state::pending::Request& request, common::service::Code code);
 
+            void add( common::message::event::service::Metric metric);
+
             explicit operator bool () const noexcept;
 
             void clear();
--- a/middleware/http/source/outbound/handle.cpp
+++ b/middleware/http/source/outbound/handle.cpp
@@ -20,6 +20,24 @@ namespace casual
    using namespace common;
    namespace http::outbound::handle
    {
+      namespace local
+      {
+         namespace
+         {
+            namespace metric::maybe
+            {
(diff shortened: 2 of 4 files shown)


middleware/http/include/http/outbound/state.h
```diff
@@ -221,6 +225,15 @@ namespace casual
...
 
+         inline void maybe_metric()
+         {
+            if( pending.requests.empty() || metric)
+            {
+               common::communication::device::blocking::send( common::communication::instance::outbound::service::manager::device(), metric.message());
+               metric.clear();
+            }
+         }
+
```
lazan: `State` should never "do anything", other than mutate it's own state. 
This to reduce complexity. Other stuff actually do stuff and mutate state to reflect what's done.
Also, with this "rule" we don't need ask our self, where "to do this stuff" -> less confusion and more coherent code.
Move this to _handle_.

There might be stuff that is "done" in a few places in other `State` -> if so, it should be removed.

lazan (changes requested):

middleware/http/include/http/outbound/state.h
```diff
@@ -17,6 +17,8 @@
 #include "common/message/service.h"
 #include "common/domain.h"
 #include "common/predicate.h"
+#include "common/communication/ipc.h"
+#include "common/communication/instance.h"
```
lazan: Remove

lazan (approved):
Looking good.

ahhud (approved):

PR #415, opened 2024-09-04
queue: correct runlevel for queue-forward-group

### LINKED ISSUE #221: casual-queue-forward-group never goes to runlevel=running
Runlevel for casual-queue-forward-group stays in runlevel startup.
It ought to be _running_.
This doesn´t affect anything other then symmetri and problem analyzing.

AlbsAgren:
Before: casual-queue-forward-group would never reach runlevel::running, remaining in startup until it was shut down.
Now: we set runlevel::running after receiving config from the queue-manager.

Resolves #221

### DIFF
--- a/middleware/queue/source/forward/handle.cpp
+++ b/middleware/queue/source/forward/handle.cpp
@@ -453,6 +453,8 @@ namespace casual
 
                         comply::to::state( state);
 
+                        state.runlevel = decltype( state.runlevel())::running;
+
                         state.multiplex.send( message.process.ipc, common::message::reverse::type( message, process::handle()));
                      };
                   }


middleware/queue/source/forward/handle.cpp
```diff
@@ -453,6 +453,9 @@ namespace casual
 
                         comply::to::state( state);
 
+                        if( state.runlevel < decltype( state.runlevel())::running)
+                           state.runlevel = decltype( state.runlevel())::running;
+
```
lazan, line 457: You could just assign it to `::running`. The (only forward) state machine will only assign if the 'state' is lower/less. Hence it can only move forward through the "states".

lazan (approved):

PR #416, opened 2024-09-10
queue: add 'signaled' error-code to queue api

### LINKED ISSUE #291: queue: return some equivalent of TPGOTSIG if queue api call fails due to receiving a signal
Give the user a clearer idea of what went wrong.

AlbsAgren:
This commit adds the 'CASUAL_QE_SIGNALED' error code to the queue C-api, to be returned when a call fails due to a signal interrupt.

Resolves #291


lazan (approved):
Looking good.

PR #417, opened 2024-09-13
Feature/1.7/rename transaction scale instances

AlbsAgren:


middleware/queue/source/forward/handle.cpp
```diff
@@ -430,7 +430,7 @@ namespace casual
 
                         state.alias = message.model.alias;
                         state.memberships = message.model.memberships;
-                        state.group_coordinator.update( message.groups);
+                        state.group_coordinator = { message.groups};
```
lazan, line 433: Looks a lot more c++:ish :)

middleware/transaction/include/transaction/manager/admin/model.h
```diff
@@ -345,7 +345,7 @@ namespace casual
          )
       };
 
-      namespace scale
+      namespace scale::resource::proxy
```
lazan, line 348: Good call to change the name.

middleware/transaction/include/transaction/manager/admin/server.h
```diff
@@ -19,9 +19,10 @@ namespace casual
          {
             constexpr std::string_view state = ".casual/transaction/state";
 
-            namespace scale
+            namespace scale::resource
             {
-               constexpr std::string_view instances = ".casual/transaction/scale/instances";
+               // TODO: rename this?
```
lazan: Yes, or rather, keep the old one also. I think we set category to _deprecated_.

middleware/transaction/source/manager/admin/server.cpp
```diff
@@ -64,8 +64,8 @@ namespace casual
                         common::service::visibility::Type::undiscoverable,
                         common::service::category::admin
                      },
-                     { service::name::scale::instances,
-                        std::bind( &local::scale::instances, std::placeholders::_1, std::ref( state)),
+                     { service::name::scale::resource::proxies,
+                        std::bind( &local::scale::resource::proxy::instances, std::placeholders::_1, std::ref( state)),
```
lazan: You could get rid of `std::bind`, which sorta has played out it's role and replaced with generic lambdas. The placeholder stuff is not something to write home abut...

lazan (changes requested):
Looking good, but I think you need to deprecate the service and create a new one with appropriate name.

middleware/transaction/source/manager/admin/server.cpp
```diff
@@ -25,32 +25,33 @@ namespace casual
...
-                        protocol >> CASUAL_NAMED_VALUE( instances);
-
-                        auto result = serviceframework::service::user( protocol, &action::resource::instances, state, std::move( instances));
-
-                        protocol << CASUAL_NAMED_VALUE( result);
-                        return protocol.finalize();
+                        return [ &state]( common::service::invoke::Parameter&& parameter)
+                        {
+                           auto protocol = serviceframework::service::protocol::deduce( std::move( parameter));
+                           auto instances = protocol.extract< std::vector< admin::model::scale::resource::proxy::Instances>>( "instances");
```
lazan, line 45: A lot cleaner.

middleware/transaction/source/manager/admin/server.cpp
```diff
@@ -59,13 +60,20 @@ namespace casual
...
                         common::service::category::admin
                      },
-                     { service::name::scale::instances,
-                        std::bind( &local::scale::instances, std::placeholders::_1, std::ref( state)),
+                     // deprecated
+                     { ".casual/transaction/scale/instances",
+                        local::scale::resource::proxy::instances( state),
                         common::service::transaction::Type::none,
                         common::service::visibility::Type::undiscoverable,
                         common::service::category::admin
```
lazan: Change to `common::service::category::deprecated` ?

lazan (approved):

middleware/transaction/source/manager/admin/server.cpp
lazan: Hmm, this would probably expose the service during `casual service --list-services`. I think we just partition all services on predicate `service.category == .admin`  -> [services, admin-services]. 
Maybe we should use predicate `services.category.starts_with( '.')` ? Sorta same thing as hidden unix files/directorys.

PR #419, opened 2024-09-13
service: refactoring service-manager - reference based -> index based

lazan:
Before: services and instances had their circular dependencies made of pointers and references. This made it a bit error prune and hard to reason about.

Now: Every service has an id, every instance has an id. The relation between them is just "a vector with ids". This is:
* a service has all ids of instances that has advertise the service
* an instance has all the ids of services that the instance has advertised.

Also, added a _disabled_ set that holds all instances that is in _prepare for shutdown_ mode. This enables us to keep track of _disabled_ instances and be able to handle a timeout properly during this mode.

Resolves #407

### DIFF
--- a/middleware/common/include/casual/container.h
+++ b/middleware/common/include/casual/container.h
@@ -22,8 +22,17 @@ namespace casual
 {
    namespace container
    {
+      namespace detail
+      {
+         namespace id
+         {
+            struct Tag{};
+            using type = common::strong::detail::integral::Type< platform::size::type, Tag, -1>;
+         } // id
+      
+      } // detail
 
-      template< typename T, typename I>
+      template< typename T, typename I = detail::id::type>
       struct Index
       {
          using value_type = T;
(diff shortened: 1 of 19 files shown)


middleware/common/include/casual/container.h
```diff
@@ -54,54 +63,91 @@ namespace casual
...
             return std::exchange( m_data[ key.value()], {});
          }
 
+         //! @returns all values
+         auto values() const
+         {
+            auto has_value = []( auto& value){ return value.has_value();};
+            auto transform = []( auto& value) -> decltype( *value) { return *value;};
+            
+            return m_data | std::ranges::views::filter( has_value) | std::ranges::views::transform( transform);
```
AlbsAgren, line 108: Cool

middleware/common/include/casual/container.h
```diff
@@ -54,54 +63,91 @@ namespace casual
...
 
-         auto begin() noexcept { return std::begin( m_data);}
-         auto end() noexcept { return std::begin( m_data);}
-         auto begin() const noexcept { return std::begin( m_data);}
-         auto end() const noexcept { return std::begin( m_data);}
-         
+            // Missing enumerate
+            // auto has_value = []( auto index, auto& value){ return value.has_value();};
+            // auto transform = []( auto index, auto& value) -> index_type { return index_type( index);};
+            // return m_data | std::views::enumerate | std::ranges::views::filter( has_value) | std::ranges::views::transform( transform);
```
AlbsAgren, line 125: Remove?

middleware/service/source/manager/state.cpp
```diff
@@ -320,69 +258,150 @@ namespace casual
...
+               if( pair.second != id)
+                  return false;
+
+               m_metrics.erase( pair.first);
+               return true;
+            });
          }
 
-         bool Service::is_discoverable() const noexcept
+         void Services::restore_routes( state::service::id::type id)
```
AlbsAgren: So this erases all routes and restores the original name of the service? Would something like restore_origin_name or similar be more telling?

test/unittest/source/test_gateway.cpp
```diff
@@ -1629,16 +1629,21 @@ namespace casual
...
          });
 
-         // expect even distribution
-         EXPECT_TRUE( domains[ "B"] == 3);
-         EXPECT_TRUE( domains[ "C"] == 3);
-         EXPECT_TRUE( domains[ "D"] == 3);
+         // for a domain to have 0 calls after 100 events:
+         // ( 2 / 3 ) ^100 ~~ 1 in 4,065611775352153e17
+         // 1 in 400'000'000'000'000'000
+         // one in 400 million billions.
```
AlbsAgren, line 1641: 😄

AlbsAgren (approved):
Neat, feels at least 56% easier to grasp than before

middleware/common/include/casual/container.h
lazan: I was thinking that when we get a conformant lib-c++ we can just uncomment.

middleware/service/source/manager/state.cpp
lazan: I concur.

test/unittest/source/test_gateway.cpp
lazan: If we run the test non stop. On average it will fail every 1.2 billion years :)

middleware/common/include/casual/container.h
AlbsAgren: Ahh, right. My bad.

PR #424, opened 2024-09-19
documentation: alias is missing for forward.groups.services/queues

lazan:
This patch:

Adds the missing documentation for queue.forward.groups.queues and documents `alias` for queue.forward.groups.services. Adds `alias` to the queue example configuration


partickl (approved):
looking good

PR #429, opened 2024-09-27
Feature/1.7/service list

lazan:
Primarily  commit: c670ee34aa03e71626e735475208e384fbb16d12 

Started to refactor `argument` to be able to have _pure flag options_, without `--flag <true|false>`, but it was too much work. I'll do it later. Kept the refactoring to use _views_ for serialize writes.


AlbsAgren (approved):
Looks good to me!

PR #430, opened 2024-09-29
common: simplifying implementation of timepoint formatting

KristianIvarsson:
Using semi modern std::chrono and std::format to format output

An earlier pull-request was based upon a branch that eventually contained merge-flaws and other weird things, so this branch is the quick fix replacement for that


lazan (approved):
A lot cleaner!

PR #432, opened 2024-10-01
gateway: honor group memberships on connections

AlbsAgren:
This obviously needs a bunch of tests, but I'm doing a draft too see what you think of the overall idea. 

Basically, by filtering out disabled connections before calculating the 'difference' we can piggyback on the existing implementation. The alternative would probably be something like putting them in the connection::Holder and moving them around as their 'enabledness' changes, but this felt way too cumbersome when I tried to implement it.

### DIFF
--- a/middleware/administration/unittest/source/cli/test_queue.cpp
+++ b/middleware/administration/unittest/source/cli/test_queue.cpp
@@ -5,6 +5,8 @@
 //!
 
 #include "common/unittest.h"
+#include "common/unittest/file.h"
+#include "common/unittest/environment.h"
 
 #include "administration/unittest/cli/command.h"
 
@@ -14,6 +16,9 @@
 #include "gateway/unittest/utility.h"
 
 #include "common/string.h"
+#include "common/sink.h"
+
+#include "sql/database.h"
 
 #include <regex>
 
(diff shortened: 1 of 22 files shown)


middleware/configuration/source/model.cpp
```diff
@@ -621,14 +621,22 @@ namespace casual
...
                entity.enabled = coordinator.enabled( entity.memberships);
             };
 
             algorithm::for_each( model.domain.servers, update_enabled);
             algorithm::for_each( model.domain.executables, update_enabled);
+
+            auto update_connections = [ &]( auto& group)
+            {
+               algorithm::for_each( group.connections, update_enabled);
+            };
```
lazan, line 636: Pretty nice with the `coordinator`. Easy and coherent with the other stuff.

middleware/configuration/source/model/transform.cpp
```diff
@@ -779,6 +781,7 @@ namespace casual
                         configuration::user::domain::gateway::inbound::Connection result;
                         result.address = value.address;
                         result.note = null_if_empty( value.note);
+                        result.memberships = value.memberships;
```
lazan: Shouldn't this be "null if empty"?

middleware/gateway/source/group/inbound/main.cpp
```diff
@@ -82,11 +84,16 @@ namespace casual
                            state.alias = message.model.alias;
                            state.limit = message.model.limit;
 
+                           auto is_enabled = []( auto& connection){ return connection.enabled;};
+                           auto [ enabled, disabled] = algorithm::stable::partition( message.model.connections, is_enabled);
+
+                           state.disabled_connections = algorithm::container::vector::create( disabled);
+
```
lazan, line 91: Ahh, right. The disabled are kept on the side, and if any of them are present, they will end upp in `change.removed`, and disconnect as before...

lazan (commented):
I think this will work. It feels a bit strange to "fake" the change algorithm, but I've convinced my self that it works, and probably is easier than the alternatives.

A few unittests missing? 😄

AlbsAgren:
Cool, I'll get on with writing some tests.

middleware/configuration/source/model/transform.cpp
AlbsAgren: Yes, good catch

middleware/administration/unittest/source/cli/test_queue.cpp
```diff
@@ -490,5 +490,66 @@ b2        external  50233  out    B
...
+         }
+
+         // verify queue is empty
+         {
+            auto capture = local::execute( R"(casual queue --list-messages a --porcelain true)");
+            EXPECT_TRUE( capture.standard.out.empty()) << CASUAL_NAMED_VALUE( capture);
+         }
+      }
+
+      TEST( cli_queue, remove_messages)
```
AlbsAgren, line 532: I wanted to write one with `--force` but couldnt figure out how to get some messages in the right state, any ideas?

lazan: Hmm, It's going to be a bit cumbersome...
* create a queue-group with a specific queuebase name.
* enqueue some messages
* shutdown the queue group (or the domain)
* open the queuebase and alter the state of the messages 
* boot the queue group
* perform tests
* prosper...

AlbsAgren: Right, thats not too bad, I'll have a go at it

middleware/administration/unittest/source/cli/test_queue.cpp
```diff
@@ -490,5 +495,124 @@ b2        external  50233  out    B
...
+            auto capture = local::execute( list_message_ids, "| xargs -I {} casual --porcelain true queue --remove-messages a {} --force | wc -l");
+            EXPECT_TRUE( std::stoi( capture.standard.out) == 10) << CASUAL_NAMED_VALUE( capture);
+         }
+
+         {
+            auto capture = local::execute( R"(casual --porcelain true queue --list-messages a | wc -l)");
+            EXPECT_TRUE( std::stoi( capture.standard.out) == 0) << CASUAL_NAMED_VALUE( capture);
+         }
+      }
+
```
lazan, line 616: Nice

lazan (approved):

PR #434, opened 2024-10-08
transaction: error log when resource fail

### LINKED ISSUE #428: TM should log error if any of commit-replies indicate error
It does not seems like we error-log if a resource commit fails. This is pretty much the worst error we could detect. Some resources got commited, some not so much...

lazan:
Error log when resource commit/rollback returns other than xa:ok.

We don't error log on prepare fails, since this could be an "expected" outcome. Resources use optimistic locking and so on.

Resolves #428

I'm also going to update queue documentation in this PR.


AlbsAgren (approved):
👍

partickl (approved):

PR #437, opened 2024-10-10
service: fix routes dissapearing on config update

### LINKED ISSUE #431: Enabling a server does not add it as an instance in routes pointing to its services
Reproduce:

Boot the following domain:
```yaml
domain:
  name: test

  groups:
    - name: my-group
      enabled: false

  servers:
      instances: 1
      memberships:
        - my-group

  services:
    -  name: casual/example/domain/name
       routes: [ my-route]
```

Try to call the route:
```
$ echo "" | casual buffer --compose | casual call --service my-route | casual buffer --extract
[xatmi:TPENOENT] TPENOENT
```

Enable the group:
```
$ casual configuration --enable-groups my-group
```

Call the route again:
```
$ echo "" | casual buffer --compose | casual call --service my-route | casual buffer --extract
[xatmi:TPENOENT] TPENOENT
```
Here one would expect the call to succeed since there is now an instance of the service to which the route points.

AlbsAgren:
This commit fixes an issue where services that did not have their configuration changed would lose their routes when the service manager received a runtime configuration update.

Resolves #431

### DIFF
--- a/middleware/service/unittest/source/manager/test_manager.cpp
+++ b/middleware/service/unittest/source/manager/test_manager.cpp
@@ -80,16 +80,6 @@ domain:
                return domain::unittest::manager( configuration::base, std::forward< C>( configurations)...);
             }
 
-            namespace call
-            {
-               auto state()
-               {
-                  return serviceframework::service::protocol::binary::Call{}( 
-                     manager::admin::service::name::state).extract< manager::admin::model::State>();
-               }
-
-            } // call
-
             namespace service
             {
                const manager::admin::model::Service* find( const manager::admin::model::State& state, const std::string& name)
@@ -262,7 +252,7 @@ domain:
 
          auto domain = local::domain();
 
-         auto state = local::call::state();
+         auto state = unittest::state();
 
          EXPECT_TRUE( local::has_services( state.services, { manager::admin::service::name::state}));
          EXPECT_FALSE( local::has_services( state.services, { "non/existent/service"}));
@@ -277,7 +267,7 @@ domain:
 
          service::unittest::advertise( { "service1", "service2"});
 
-         auto state = local::call::state();
+         auto state = unittest::state();
 
          ASSERT_TRUE( local::has_services( state.services, { manager::admin::service::name::state, "service1", "service2"}));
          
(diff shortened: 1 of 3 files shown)


middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1226,5 +1226,43 @@ namespace casual
          }
       }
 
+      TEST( service_manager, runtime_update_routes)
```
AlbsAgren, line 1219: This is maybe a bit too specific...

lazan: In what regard?

middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1226,5 +1226,43 @@ namespace casual
...
+
+         auto domain = local::domain( R"(
+domain:
+   services:
+      -  name: a
+         routes: [ b]
+)");
+
+         {
+            auto state = local::call::state();
```
lazan: Hmm, SM should have a `state()` function in `unittest/utillity.h` that we should use?

middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1226,5 +1226,43 @@ namespace casual
...
+         auto wanted = local::configuration::load( local::configuration::base, R"(
+domain:
+   services:
+      -  name: a
+         routes: [ b]
+      -  name: c
+         routes: [ d]
+)");
+
+         casual::configuration::model::transform( casual::domain::unittest::configuration::post( casual::configuration::model::transform( wanted)));
```
lazan: Why do you transform the result of `post` and not bother with the return?

middleware/service/unittest/source/manager/test_manager.cpp
lazan: Hence, remove the `local::call::state()` function -> alter the stuff in this file that uses it.

middleware/service/unittest/source/manager/test_manager.cpp
AlbsAgren: I wondered if maybe it should test that the whole state has been correctly updated, rather than just the routes, but I'm not sure.

lazan: > I wondered if maybe it should test that the whole state has been correctly updated, rather than just the routes, but I'm not sure.

I think it's better to be finer then coarser.

lazan (approved):

PR #439, opened 2024-10-15
buffer: simplify the handling of the special 'inbound' buffer

### LINKED ISSUE #411: Make sure to ensure mecanism for deallocate incoming buffers
Checkout `casual::buffer::field::local::Buffer::m_inbound` (CFIELD) that have a mechnism that doesn't exists in CSTRING, COCTET and CORDER

lazan:
This patch simplifies keeping track of the special 'inbound' buffer. We now keep track of 'inbound' for all our current known buffers.

Resolves #411

### DIFF
--- a/middleware/common/include/common/buffer/type.h
+++ b/middleware/common/include/common/buffer/type.h
@@ -30,7 +30,6 @@ namespace casual
 {
    namespace common::buffer
    {
-
       namespace handle
       {
          namespace detail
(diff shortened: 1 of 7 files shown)


middleware/common/include/common/buffer/type.h
```diff
@@ -184,13 +188,15 @@ namespace casual
          inline buffer::handle::type handle() const noexcept { return payload.handle();}
          inline buffer::handle::mutate::type handle() noexcept { return payload.handle();}
 
-         //inline friend bool operator == ( const Buffer& lhs, buffer::handle::type rhs) { return lhs.handle() == rhs;}
+         bool inbound() const { return m_inbound;}
```
KristianIvarsson, line 191: could be `noexcept`

KristianIvarsson (approved):
Looking neat and slim and good and beautiful

middleware/common/include/common/buffer/type.h
lazan: Yeah, but I don't think noexcept improve that much. Besides move semantics and such. I also believe that the compiler can deduce inline getters as conceptual noexcept. With noexcept the compiler has to add guards against exceptions.
I'm not so sure about sprinkle noexcept all over any more...

PR #444, opened 2024-10-17
Feature/1.7/unittest

lazan:


middleware/gateway/include/gateway/manager/admin/model.h
```diff
@@ -82,6 +83,7 @@ namespace casual
                   case Runlevel::connecting: return "connecting";
                   case Runlevel::pending: return "pending";
                   case Runlevel::connected: return "connected";
+                  case Runlevel::disconnecting: return "disconnecting";
```
KristianIvarsson, line 86: Not related to this commit, but we should have some kind of macro (or such) for an enum-to-string-thing

middleware/gateway/source/manager/admin/cli.cpp
```diff
@@ -154,6 +154,7 @@ namespace casual
                            case Enum::connecting: return 10;
                            case Enum::pending: return 7;
                            case Enum::connected: return 9;
+                           case Enum::disconnecting: return 13;
```
KristianIvarsson, line 157: WIth some kind of macro thing and an enum-to-string-thing this could somehow be automated (and maybe constexpr'd) (still not related to this commit)

KristianIvarsson (approved):

middleware/gateway/source/manager/admin/cli.cpp
lazan: This particular thing should be: 
```cpp
auto width( const model::Connection& value, const std::ostream&) const
{
    return std::ssize( description( value.runlevel)); 
}
```

middleware/gateway/include/gateway/manager/admin/model.h
lazan: Well, it depends on the _macro_. But I don't think the benefits are that good. It's not hard to write the `description`, and the compiler will tell you if you forgot values, or values are added over time.
Hence, the benefits are only convenience at best, and not to help/force doing the correct thing, since this is covered by the compiler.
But you might think of a _macro_ that does some "magical introspection" stuff?

ahhud (approved):

AlbsAgren (approved):

PR #448, opened 2024-10-22
discovery: remove invalid sanity check

### LINKED ISSUE #443: discovery invalid-semantic error
When running unittest, suspiciously many errors with invalid-semantic

```
casual-domain-discovery||||error|[casual:invalid-semantics] unexpected inflight: 0
````

Is this a "false negative"?

AlbsAgren:
Before: The accumulate functionality in casual-domain-discovery had a sanity check that would log an error invalid-semantics error if there were less pending than requests in-flight. This check did not account for the fact that topolygy requests are always accumulated, making this situation 'valid'.

Now: The check has been removed.

Resolves #443

### DIFF
--- a/middleware/domain/source/discovery/state.cpp
+++ b/middleware/domain/source/discovery/state.cpp
@@ -118,14 +118,9 @@ namespace casual
                   {
                      auto pending = heuristic.pending_requests();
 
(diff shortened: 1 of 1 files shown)


lazan (approved):
Looks good

PR #455, opened 2024-11-04
service: discovery fetch known only discoverable

### LINKED ISSUE #454: service discovery fetch known should not "know" undiscoverable services
If a service is _undiscovarable_ we should not include it in discoveries. `domain-discovery` ask SM for known services -> SM should only reply with known "external" discoverable services.

lazan:
This patch:
Excludes services that is _undiscoverable_ for the discovery::fetch::known::Reply

Resolves #454


partickl (approved):
looking good

ahhud (approved):

PR #457, opened 2024-11-07
build: simplify install script for http-inbound

AlbsAgren:
This is basically all that I felt was doable in 1.7, do you agree?

### DIFF
--- a/thirdparty/nginx/nginx.conf
+++ b/thirdparty/nginx/nginx.conf
@@ -14,7 +14,6 @@ events {
 }
 
 http {
-    include       mime.types;
     default_type  application/octet-stream;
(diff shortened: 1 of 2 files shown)


lazan:
We "solved" two of the six:ish improvements?

AlbsAgren:
--prefix has to stay, else nginx builds to `whatever_outputdir_we_chose/usr/local/nginx`. And the other one's will break stuff. I can do them, but we have to at least be aware of the consequences.

lazan (approved):

lazan:
Do you modify the issue, or write a new one with the remaining stuff?

ahhud (approved):

PR #461, opened 2024-11-13
Feature/1.7/potential stale transaction

### LINKED ISSUE #440: Resurrection of rolled backed  transactions could happen
Scenario: 

* A starts `trid X`
* A -> B
* B has timeout and linger contract
* B timeouts and SM gives error reply to A
* A rollback `trid X`
* TM issue rollback to all involved resources to `trid X`
* B -> C
* C is bound to resource Y, and  involves `trid X` 
* TM associate `trid X` with resource Y, _resurrect_ `trid X` 

`trid X` is now a _stale_ trid, and will not be rolled back by `casual`

lazan:


middleware/transaction/unittest/source/test_manager.cpp
```diff
@@ -2012,7 +2012,97 @@ namespace casual
...
+
+
+         // send the potentially stale message, this will trigger a rollback.
+         {
+            common::message::transaction::potential::Stale message{ process::handle()};
+            message.gtrid = common::transaction::id::range::global( trid);
+            local::send::tm( message);
+         }
+
+         local::handle_rollback_request( resources, trid);
```
AlbsAgren: How would this test fail? By hanging indefinitely?

AlbsAgren (approved):
Looks good. I think the SM tests are particularly useful for understanding the actual scenario

middleware/transaction/unittest/source/test_manager.cpp
lazan: Yes.

lazan: Do you think we need to no-block pull just in case?

AlbsAgren: I don't know about need, but I'd argue it's somewhat more clear what went wrong if we get a "normal" test failure.

OlofGullnas:
A lot of code to look at :-)  My feeling is that only looking at changes is not very meningsful for me as I am not familiar with the TM implementation.

I built this branch and tried the unit tests!

UPDATE:
Seems as if there was an inconsistent build. Tests transaction_manager.* runs after clean and rebuild, so the issues mentioned below seems to be a non-issue!
 END UPDATE

Consistently hangs in the test:
transaction_manager.begin_commit_transaction__1_resources_involved__expect_one_phase_commit_optimization

If I disable that test with GTEST_FILTER it hangs in:
transaction_manager.begin_commit_transaction__1_resources_involved__2_times___expect_one_phase_commit_optimization

If I disable that test also it hangs in:
transaction_manager.begin_commit_transaction__1_resources_involved__xa_XA_RBDEADLOCK___expect__TX_HAZARD

I have no tried disabling that test also, but there seems to be some problem...

As usual I am using a debug build on a 1 cpu/1 core virtual machine.

Just to make sure I do not have problems with an inconsistent build I will do a "clean" and rebuild. It takes some time on this system... The same directory tree was used to build feature/1.7/main before switching branch...

The build in 1.7/main did not have problems with theese transaction_manager tests.

lazan:
> A lot of code to look at :-) My feeling is that only looking at changes is not very meningsful for me as I am not familiar with the TM implementation.
> 
> I built this branch and tried the unit tests!
> 
> UPDATE: Seems as if there was an inconsistent build. Tests transaction_manager.* runs after clean and rebuild, so the issues mentioned below seems to be a non-issue! END UPDATE
> 
> Consistently hangs in the test: transaction_manager.begin_commit_transaction__1_resources_involved__expect_one_phase_commit_optimization
> 
> If I disable that test with GTEST_FILTER it hangs in: transaction_manager.begin_commit_transaction__1_resources_involved__2_times___expect_one_phase_commit_optimization
> 
> If I disable that test also it hangs in: transaction_manager.begin_commit_transaction__1_resources_involved__xa_XA_RBDEADLOCK___expect__TX_HAZARD
> 
> I have no tried disabling that test also, but there seems to be some problem...
> 
> As usual I am using a debug build on a 1 cpu/1 core virtual machine.
> 
> Just to make sure I do not have problems with an inconsistent build I will do a "clean" and rebuild. It takes some time on this system... The same directory tree was used to build feature/1.7/main before switching branch...
> 
> The build in 1.7/main did not have problems with theese transaction_manager tests.

Does all the unittest work on `feature/1.7/main`?

I don't think we'll merge this patch right now. It's to complex and still doesn't work in all corner cases.

OlofGullnas:
Hm...
My impression is that unit tests works better in 1.7/main.  The most common problem there are sporadic hangs in some test steps, but not very frequently. It takes ¨35 minutes to do a total rebuild so witching back and forth takes some time (I could set up separate trees...)

I did some more tests on this branch  and got 

100% hangs in test service_manager.service_a_execution_timeout_duration_1ms__advertise_a__reserve_a__lookup_a__expect_lookup_timeout
(The other service_manager.* tests has not failed in my tests when I filter out the failing one)

apprx 30% chance of failure in  cli_transaction.pending_resource_proxies
 [ RUN      ] cli_transaction.pending_resource_proxies
/export/home/guo/c_1.7/casual-1.7/middleware/administration/unittest/source/cli/test_transaction.cpp:283: Failure
Value of: string::from< double>( output[ 11]) > 0.01
  Actual: false
Expected: true

Also just got a failure in (this is with the above tests filtered out):
[ RUN      ] cli_queue.list_queue_instances
/export/home/guo/c_1.7/casual-1.7/middleware/administration/unittest/source/cli/test_queue.cpp:492: Failure
Value of: std::regex_match( rows.at( 1), std::regex{ R"(b2\|external\|[0-9]+\|out\|B)"})
  Actual: false
Expected: true
[  FAILED  ] cli_queue.list_queue_instances (390 ms)

This is an error I have also seen in feature/1.7/main and that has an issue reported.

I will do a rebuild of 1.7/main and run some tests there, but not know. Tonight or tomorrow...

OlofGullnas:
I tried the feature/1.7/main branch
On the first execution the tests failed in cli_queue.list_queue_instances, I removed that test with GTEST_FILTER.

On 2:nd execution I got:
`[ RUN      ] test_service.multiplex_call_to_service__kill_server__expect_multiplex_detect_ipc_device_is_gone
/export/home/guo/c_1.7/casual-1.7/test/unittest/source/test_service.cpp:434: Failure
Value of: ! coordinator.empty()
  Actual: false
Expected: true
/export/home/guo/c_1.7/casual-1.7/test/unittest/source/test_service.cpp:459: Failure
Value of: error_count > 0
  Actual: false
Expected: true
[  FAILED  ] test_service.multiplex_call_to_service__kill_server__expect_multiplex_detect_ipc_device_is_gone (182 ms)`

I executed that test in isolation and it failed again on the 8:th run. But then I could run it in isolation 50 times without failure.
I then went back to running the complete suite (except the cli_queue.list_queue_instrances).
This failed with:

`[ RUN      ] gateway_manager.A_B__pending_call_A_to_B__shutdown_B__send_implicit_update_to_A_outbound__expect_direct_reply__not_over_to_inbound
unknown file: Failure
C++ exception with description "vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)" thrown in the test body.
[  FAILED  ] gateway_manager.A_B__pending_call_A_to_B__shutdown_B__send_implicit_update_to_A_outbound__expect_direct_reply__not_over_to_inbound (285 ms)
`

This is a somewhat unusual type of failure :-)  I could run the test in isolation 50 times without failure.
The back to complete test suite... And it runs to completion (with cli_queue.list_queue_instances filtered out) for 10 executions. Then I stopped....

My impression (!!) is that the probability of failure is reduced in repeated runs. A possibility is that more stuff being in memory (e.g. file system cache) affects timing to reduce the risk of failure. I use debug bioöld on a single core CPU which may be a contributing factor.

lazan:
I'll remove this PR, since we're not shipping it yet, and in this form.

@OlofGullnas Could you write specific issues for the testcases that fails for you?

PR #467, opened 2024-11-21
queue: fix forwards ignoring group memberships

### LINKED ISSUE #465: Enabling/disabling groups have no effect on member forwards

AlbsAgren:
This commit fixes an issue where forwards would not receive configuration updates when their group memberships changed. This made it so that enabling or disabling groups that a forward was dependent on would not affect the forward.

Resolves #465

### DIFF
--- a/middleware/queue/include/queue/forward/state.h
+++ b/middleware/queue/include/queue/forward/state.h
@@ -20,8 +20,6 @@
 #include "common/communication/select.h"
 #include "common/communication/ipc/send.h"
 
-#include "configuration/group.h"
-
 #include <vector>
 #include <string>
 #include <iosfwd>
@@ -446,7 +444,7 @@ namespace casual
 
          std::vector< std::string> memberships;
 
-         configuration::group::Coordinator group_coordinator;
+         bool enabled;
 
          //! we're done when we're in shutdown mode
          //! and all forwards has no concurrent stuff in flight.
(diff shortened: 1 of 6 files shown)


ahhud (approved):

middleware/queue/include/queue/forward/state.h
```diff
@@ -446,7 +444,7 @@ namespace casual
 
          std::vector< std::string> memberships;
 
-         configuration::group::Coordinator group_coordinator;
+         bool enabled;
```
lazan, line 447: Initialize the bool

lazan: Can the whole forward group be enabled/disabled?

AlbsAgren: Yes, which is asymmetric with gateway I suppose. I think forward was the first one we did.

lazan: But shouldn't we shutdown the group? Or is to complicated?

AlbsAgren: Yeah, I'm not sure why I did it like this. I imagine we had the same discussion back then, but I can't remember.

lazan: I don't remember us talking about having a _disabled_ process running. I do remember us talking about keeping an _enable indicator_ for each forward state that the queue-manager has.

But then again... I don't really trust my memory :)

lazan: I think you should change the samantics to shutdown a disabled group. You game?

AlbsAgren: Aye

PR #475, opened 2024-12-02
queue: default queue-group alias missing

### LINKED ISSUE #474: We don't generat a default queue group alias when user does not provide one

lazan:
Before: We didn't set a default queue-group alias if the user didn't provide one. This could give a whole bunch of problems. If you had two _unaliased_ queue group, both of them tries to create the queuebase file at the same path, since queuebase path is based on alias.

Now: we set a default `group`, `group.2`... `group.N`, as we do elsewhere.

resolves #474


AlbsAgren (approved):
Nice

PR #477, opened 2024-12-03
queue: reset queue group metric on startup

### LINKED ISSUE #469: Make sure we clean metrics on startup for casual-queue
We persist metrics, which nobody wants as it turns out. 
I'm convinced that it's easiest to just reset the metric on startup, and keep using sqlite for bookkeeping.

lazan:
The metric for totally enqueued and dequeued committed messages is kept in the `qeueue` table. Hence the metrics are persistent (if the queue-group is persistent).

* This behavior is not coherent with the rest of casual metrics
* Users has requested metrics not to be persistent

This patch resets the metrics for enqueued/dequeued to `0` during startup.

resolves #469

### DIFF
--- a/middleware/queue/source/group/queuebase/statement.cpp
+++ b/middleware/queue/source/group/queuebase/statement.cpp
@@ -293,5 +293,15 @@ WHERE queue = :queue AND state = 2 AND timestamp > :timestamp AND available < :a
          return result;
       }
 
+      namespace metric
+      {
+         std::string_view reset()
+         {
+            return R"( 
+            UPDATE queue
+            SET metric_dequeued = 0, metric_enqueued = 0; )";
+         }
+      } // metric
+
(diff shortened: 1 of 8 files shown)


middleware/queue/source/group/queuebase/statement.cpp
```diff
@@ -293,5 +293,15 @@ WHERE queue = :queue AND state = 2 AND timestamp > :timestamp AND available < :a
...
 
+      namespace metric
+      {
+         std::string_view reset()
+         {
+            return R"( 
+            UPDATE queue
+            SET metric_dequeued = 0, metric_enqueued = 0; )";
+         }
+      } // metric
```
lazan, line 304: CoPilot suggested exactly this... _What a time to be alive!_ :)

ahhud (approved):

PR #481, opened 2024-12-05
Feature/1.7/casual file

### LINKED ISSUE #13: casual_file

KristianIvarsson:
This is targeting (earliest) 1.8 and just a heads up for an early review to identify flaws at an early stage, i.e. it's not yet ready (and contain some fluffy stuff that might need an overhaul)

### DIFF
--- a/middleware/common/include/common/message/type.h
+++ b/middleware/common/include/common/message/type.h
@@ -355,6 +355,10 @@ namespace casual
          internal_dump_state = INTERNAL_BASE,
          internal_configure_log,
 
+         FILE_BASE = 12000,
+         file_reserve_request = FILE_BASE,
+         file_reserve_reply,
+
          UNITTEST_BASE = 10000000, // avoid conflict with real messages
          unittest_message = UNITTEST_BASE,
       };
--- a/middleware/file/documentation/file.development.md
+++ b/middleware/file/documentation/file.development.md
@@ -1,0 +1,75 @@
+# casual file development
+
+`casual-file` is a built in mechanism to manipulate files within (XA) transactions and thus it requires a transaction.
+
+`casual-file` requires no configuration.
+
+## isolation
+
+`casual::file::blocking::reserve()` is a blocking operation so the caller have to wait for potentiall other transactions using the same path.
+
+`casual::file::non::blocking::reserve()` returns immediately if the resource (path) is used by other transactions.
+
+Multiple access to the same (original) path within the same transaction if of course possible.
+
+## notes
+
+`casual-file` implies some overhead (apart from communication) compared to plain file operations in order to be able to commit and rollback content (maybe needless to write).
+
+## samples
+
+### create a file
+
+```c++
+#include <api/file.h>
+...
+   const auto path = casual::file::reserve( the_path);
+
+   std::ofstream{ path} << data;
+...
+```
+
+### update a file
+
+```c++
+#include <api/file.h>
+...
+   const auto path = casual::file::reserve( the_path);
+
+   std::ofstream{ path} << data;
+...
+```
+
+### read a file
+
+```c++
+#include <api/file.h>
+...
+   const auto path = casual::file::reserve( the_path);
+
+   std::ifstream{ path} >> data;
+...
+```
+
+### remove a file
+
+```c++
+#include <api/file.h>
+...
+   const auto path = casual::file::reserve( the_path);
+
+   std::filesystem::remove( path);
+...
+```
+
+### rename a file
+
+```c++
+#include <api/file.h>
+...
+   const auto source = casual::file::reserve( the_source_path);
+   const auto target = casual::file::reserve( the_target_path);
+
+   std::filesystem::rename( source, target);
+...
+```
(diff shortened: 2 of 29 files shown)


middleware/common/include/common/message/type.h
```diff
@@ -355,6 +355,10 @@ namespace casual
          internal_dump_state = INTERNAL_BASE,
          internal_configure_log,
 
+         FILE_BASE = 12000,
+         file_reserve_request = FILE_BASE,
+         file_reserve_reply,
+
```
lazan, line 361: I think reserve is better than acquire, and easier to spell :)

middleware/file/include/file/api/file.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace casual
+{
+   namespace file
+   {
+      inline namespace v1  
+      {
+         //! reserve a path with blocking wait
+         [[nodiscard]] auto acquire( std::filesystem::path path) -> std::filesystem::path;
+
```
lazan, line 23: I like `reserve` better.

middleware/file/include/file/message.h
```diff
@@ -0,0 +1,111 @@
...
+
+                  if( transaction)
+                  {
+                     // there's a transaction so make sure to trigger an interaction with the TM
+                     transaction.external();
+                  }
+                  else
+                  {
+                     common::code::raise::error( common::code::casual::preconditions, "no transaction");
+                  }
```
lazan: I don't think _message_ should be responsible for the validity of it's content. I think you should move this to some other place. And get rid of the include of transaction context and such.

middleware/file/include/file/resource/logic.h
```diff
@@ -0,0 +1,32 @@
...
+
+   using Commit   = common::message::transaction::resource::commit::Request;
+   using Rollback = common::message::transaction::resource::rollback::Request;
+   using Exit     = common::message::event::process::Exit;
+
+   using Shutdown = common::message::shutdown::Request;
+
+   void reserve(  State& state, Reserve request);
+   void commit(   State& state, const Commit& request);
+   void rollback( State& state, const Rollback& request);
```
lazan: I hate it! :) why the `ws` ?. It's subjective of course, but we don't try to line upp stuff elsewhere, and it ads work when you want to add/change stuff in the future.

middleware/file/source/resource/logic.cpp
```diff
@@ -0,0 +1,301 @@
...
+   namespace local
+   {
+      namespace
+      {
+         auto temporary( const common::transaction::ID& transaction, const std::filesystem::path& path)
+         {
+            std::filesystem::path result{ path};
+            result += ".casual.";
+            result += (std::ostringstream{} << transaction).str();
+            return result;
```
lazan: You could use `common::string::compose( ...)`.
Hmm,  shouldn't it be enough with just the `gtrid`... Or do we need to keep track of the branch for the _temporary_. We only need the transaction part of the filename if the power goes?

middleware/file/unittest/source/resource/utility.cpp
```diff
@@ -0,0 +1,71 @@
...
+      }
+
+
+      namespace acquire
+      {
+         namespace local
+         {
+            namespace device
+            {
+               namespace
```
lazan, line 37: Have the unnamed ns directly under _local_.

middleware/common/include/common/code/file.h
lazan, line 1: Hmm, I don't think we ever need the _file code_ in other parts than within file it self. I suggest you move this `code` to `middleware/file`

middleware/file/documentation/file.development.md
lazan, line 1: Nice!

lazan (commented):
Did a first _read through_, will do another soon.

middleware/file/source/resource/logic.cpp
KristianIvarsson: Use `common::string::compose()`for what ? The entire path or just the postfix ? We're composing a _path_ and not a _string_, but yes, it might be useful for the postfix

middleware/file/unittest/source/resource/utility.cpp
KristianIvarsson: What is the rational for that ?

middleware/file/source/resource/logic.cpp
KristianIvarsson: Do you mean gtrid for just the file naming convention or that the whole resourse manager should keep track of gtrid's only ?

lazan: "for what" = "for that" :) 
I think we need to keep track of the whole trid. We have to treat different bquals as if the whole trid was different.

middleware/file/unittest/source/resource/utility.cpp
lazan: * consistency 
* When someone later ads stuff directly under local it might not get internal linkage
* IDE:s gets confused. Maybe this is not an issue any more, but it certainly was a problem.

PR #482, opened 2024-12-09
transaction: reopen resource on resource error

### LINKED ISSUE #476: Resource proxy: attempt to reopen on XAER_RMERR
Some resources (IBM MQ specifically) cannot recover if the connection goes down for a while before being restored, causing all interactions to fail with XAER_RMERR. This can be mitigated by reopening the resource, like we do on XAER_RMFAIL. 
While this is a bit of a hack, we've already crossed that bridge when we introduced this behavior on receiving RMFAIL.

AlbsAgren:
This commit changes the resource proxies to attempt to reopen the resource on XAER_RMERR, like they previously only did for XAER_RMFAIL. This allows the proxies to 'self-heal' in some situations where a resource temporarily becomes unavailable.

Resolves #476

### DIFF
--- a/middleware/common/include/common/algorithm/compare.h
+++ b/middleware/common/include/common/algorithm/compare.h
@@ -14,10 +14,14 @@ namespace casual
       template< typename V, typename... Vs>
       constexpr bool any( V&& value, Vs&&... values)
       {
-         auto equal = []( auto&& l, auto& r){ return l == r;}; // to mitigate g++ 9.3.1 bug
-         return ( equal( value, values) || ... ); 
+         return ( ( value == values) || ... ); 
       }
 
+      template< typename V, typename... Vs>
+      constexpr bool none( V&& value, Vs&&... values)
+      {
+         return ( ( value != values) && ... );
+      }
 
    } // common::algorithm::compare
 } // casual
--- a/middleware/common/source/transaction/resource.cpp
+++ b/middleware/common/source/transaction/resource.cpp
@@ -10,6 +10,7 @@
 #include "common/transaction/resource.h"
 #include "common/transaction/id.h"
 #include "common/environment/expand.h"
+#include "common/algorithm/compare.h"
 
 #include "common/log/line.h"
 #include "common/flag.h"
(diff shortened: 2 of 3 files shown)


middleware/common/source/transaction/resource.cpp
```diff
@@ -238,7 +238,7 @@ namespace casual
       {
          auto result = functor();
 
-         if( result != code::xa::resource_fail)
+         if( algorithm::none_of( std::vector{ code::xa::resource_fail, code::xa::resource_error}, predicate::value::equal( result)))
```
lazan: I don't like that we create a vector each time we check. Perhaps:

```cpp
if( ! algorithm::compare::any( result, code::xa::resource_fail, code::xa::resource_error))
     return result;
     
 ```
 
 Or don't negate it and do the reopen part within the if block.

**Review by lazan — CHANGES_REQUESTED** [review]

File: middleware/common/source/transaction/resource.cpp (hunk 1 above)
Comment (lazan) [reply middleware/common/source/transaction/resource.cpp]: Maybe we want a `algorithm::compare::none( ...)`, it might be more intuitive?

File: middleware/common/include/common/algorithm/compare.h [hunk 2]
```diff
@@ -18,6 +18,12 @@ namespace casual
          return ( equal( value, values) || ... ); 
       }
 
+      template< typename V, typename... Vs>
+      constexpr bool none( V&& value, Vs&&... values)
+      {
+         auto not_equal = []( auto&& l, auto& r){ return l != r;};
+         return ( not_equal( value, values) && ... );
```
lazan: While we're at it, we should remove the hack for the g++ 9 bug. Just compare directly in the fold expression.
`return ( ( value != values) && ...);`

Also for the `any( ...)` above.

lazan (approved):

middleware/common/include/common/algorithm/compare.h
lazan: I have verified that it is fixed on g++ 13.2 (probably a lot earlier)

middleware/common/include/common/algorithm/compare.h
```diff
@@ -14,10 +14,14 @@ namespace casual
...
-         auto equal = []( auto&& l, auto& r){ return l == r;}; // to mitigate g++ 9.3.1 bug
-         return ( equal( value, values) || ... ); 
+         return ( ( value == values) || ... ); 
       }
 
+      template< typename V, typename... Vs>
+      constexpr bool none( V&& value, Vs&&... values)
+      {
+         return ( ( value != values) && ... );
+      }
```
lazan, line 24: Ahh, a lot cleaner. It's pretty sweet with fold expressions...

PR #483, opened 2024-12-11
cli: better error if lookup fails on (en|de)queue

### LINKED ISSUE #463: Confusing error when en-/dequeueing from a nonexistent/disabled queue
You get `[casual:communication-unavailable] ipc - errno: [generic:No such file or directory]: communication-unavailable` which doesn't do a very good job explaining what went wrong.

AlbsAgren:
Before: if the user tried to en/dequeue from a nonexisting/disabled queue with the cli they would receive a cryptic error message about "communication unavailable".
Now: we return "no_queue" if the lookup fails which is clearer and more in line with the c api as well.

Resolves #463

### DIFF
--- a/middleware/queue/source/api/queue.cpp
+++ b/middleware/queue/source/api/queue.cpp
@@ -38,6 +38,11 @@ namespace casual
             namespace
             {
 
+               bool is_error( common::code::queue code)
+               {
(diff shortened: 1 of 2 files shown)


ahhud (approved):

lazan (approved):
Looking good.

PR #488, opened 2024-12-19
service: always send metric event on server exit

### LINKED ISSUE #486: SM does not send metric event when instance exit/cores
When SM detect that an instance has exited/cored it sends an error-reply to caller, but does not send metric-event. Hence, we're missing metric for service calls that cores.

AlbsAgren:
Before: if a server exited during a call and wasnt able to send an ACK, there would be no metric for the cancelled service call. Now: we always send a metric event if a server exits.

Resolves #486

### DIFF
--- a/middleware/service/unittest/source/manager/test_manager.cpp
+++ b/middleware/service/unittest/source/manager/test_manager.cpp
@@ -1169,5 +1169,54 @@ domain:
          }
       }
 
+      TEST( service_manager, callee_dies_during_service_call__expect_metric)
+      {
+         common::unittest::Trace trace;
+
+         auto domain = local::domain();
+
+         // a fake process that advertises some arbitrary service
+         auto callee_inbound = common::communication::ipc::inbound::Device{};
+         auto callee = common::process::Handle{ common::strong::process::id{ -5}, callee_inbound.connector().handle().ipc()};
+         service::unittest::advertise( { "a"}, callee);
+
+         // reserve the service
+         {
+            auto service = common::service::Lookup{ "a"}();
+            EXPECT_TRUE( service.service.name == "a");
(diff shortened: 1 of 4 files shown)


middleware/service/unittest/source/manager/test_manager.cpp
```diff
@@ -1169,5 +1169,53 @@ namespace casual
...
+            message.state.reason = decltype( message.state.reason)::core;
+
+            common::communication::device::blocking::send( 
+               common::communication::instance::outbound::service::manager::device(),
+               message);
+         }
+
+         // we have to resubscribe since DM thinks we died
+         common::event::subscribe( common::process::handle(), { event.type()});
+
```
lazan: This seems like it could be a timing glitch?
And why would DM think we've died?

lazan: Ahh, my bad. It's the SM that handles all service metric events. 
Hmm. You could subscripe to the event with a dedicated ipc-device, with a pid != our pid.

lazan: I think you need to have a dedicated _unittest process_ (pid, ipc) that subscribes. Or, it might be easier to let the _unittest process_ advertise the service.

AlbsAgren: I had another go at it, please take a look. It gets a bit gnarly when gathering as much information as possible for the metric...

lazan (approved):

middleware/service/unittest/source/manager/test_manager.cpp
lazan: Looking good, I cant see any timing issues now.

PR #489, opened 2024-12-22
File mechanism wants to be merged to main

### LINKED ISSUE #13: casual_file

KristianIvarsson:
The basic functionality is ready and done

Probably some CLI features will be added later on

### DIFF
--- a/middleware/file/documentation/file.development.md
+++ b/middleware/file/documentation/file.development.md
@@ -1,0 +1,102 @@
+# casual file development
+
+`casual-file` is a built in mechanism to manipulate files within (XA) transactions and thus it requires a transaction.
+
+`casual-file` requires no configuration.
+
+## isolation
+
+`casual::file::blocking::reserve()` is a blocking operation so the caller have to wait for potentiall other transactions using the same path.
+
+`casual::file::non::blocking::reserve()` returns immediately if the resource (path) is used by other transactions.
+
+Multiple access to the same (original) path within the same transaction if of course possible.
+
+## notes
+
+`casual-file` implies some overhead (apart from communication) compared to plain file operations in order to be able to commit and rollback content (maybe needless to write).
+
+## samples
+
+`tx_begin()`and `tx_commit()` are used in the these examples just to explicitly clarify that the operations must be used with transactions, but of course there could already be an existing transaction and thus no need to start one explicitly.
+
+### create a file
+
+```c++
+#include <api/file.h>
+...
+   tx_begin();
+...
+
+   const auto path = casual::file::reserve( the_path);
+   std::ofstream{ path} << data;
+
+...
+   tx_commit();
+...
+```
+
+### update a file
+
+```c++
+#include <api/file.h>
+...
+   tx_begin();
+...
+
+   const auto path = casual::file::reserve( the_path);
+   std::ofstream{ path} << data;
+
+...
+   tx_commit();
+...
+```
+
+### read a file
+
+```c++
+#include <api/file.h>
+...
+   tx_begin();
+...
+
+   const auto path = casual::file::reserve( the_path);
+   std::ifstream{ path} >> data;
+
+...
+   tx_commit();
+...
+```
+
+### remove a file
+
+```c++
+#include <api/file.h>
+...
+   tx_begin();
+...
+
+   const auto path = casual::file::reserve( the_path);
+   std::filesystem::remove( path);
+
+...
+   tx_commit();
+...
+```
+
+### rename a file
+
+```c++
+#include <api/file.h>
+...
+   tx_begin();
+...
+
+   const auto source = casual::file::reserve( the_source_path);
+   const auto target = casual::file::reserve( the_target_path);
+   std::filesystem::rename( source, target);
+
+...
+   tx_commit();
+...
+```
--- a/middleware/file/include/file/api/c/file.h
+++ b/middleware/file/include/file/api/c/file.h
@@ -1,0 +1,39 @@
+/** 
+ ** Copyright (c) 2015, The casual project
+ **
+ ** This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ **/
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+/* error codes
+ */
+#define CASUAL_FE_OK 0
+#define CASUAL_FE_BUSY 1
+#define CASUAL_FE_ERROR 2
+#define CASUAL_FE_SIGNAL 3
+
+/* @returns the error code associated last error */
+int casual_file_last_error_code();
+
+/* @returns the error text associated last error */
+extern const char* casual_file_last_error_text();
+
+
+/* @returns the reserved path if success and null on error */
+extern char* casual_file_blocking_reserve( const char* path);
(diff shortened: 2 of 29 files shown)


middleware/file/documentation/file.development.md
```diff
@@ -0,0 +1,75 @@
...
+
+```c++
+#include <api/file.h>
+...
+   const auto path = casual::file::reserve( the_path);
+
+   std::filesystem::remove( path);
+...
+```
+
```
lazan, line 86: Shouldn't we include tx_begin(), tx_commit() in these examples? Just to make it clear to the reader that transaction is the major part of the whole file-manager...

middleware/file/include/file/api/c/file.h
```diff
@@ -0,0 +1,39 @@
...
+/* @returns the reserved path if success and null on error */
+extern char* casual_file_blocking_reserve( const char* path);
+
+/* @returns the reserved path if success and null on error or if path is busy */
+extern char* casual_file_non_blocking_reserve( const char* path);
+
+/* releases the memory allocated during acquire or attempt */
+extern void casual_file_release( char* path);
+
+#ifdef __cplusplus
```
lazan, line 37: Pretty good. I think this is more than good enough. 
I would argue that a _flag_ to `casual_file_reserve` would be a more _posix style_ api. And, more importantly, opens up future non intrusive features.
But I'm fine as is.

middleware/file/include/file/code.h
lazan, line 1: Nice

middleware/file/include/file/factory.h
```diff
@@ -0,0 +1,57 @@
...
+namespace casual
+{
+   namespace file::factory
+   {
+
+      namespace create
+      {
+         namespace blocking
+         {
+            inline auto request( std::filesystem::path path)
```
lazan: Why are these inline? Any particular reason?

middleware/file/include/file/manager/admin/model.h
```diff
@@ -0,0 +1,53 @@
...
+         struct State
+         {
+            std::vector< Request> working;
+            std::vector< Request> pending;
+
+            CASUAL_CONST_CORRECT_SERIALIZE(
+               CASUAL_SERIALIZE( working);
+               CASUAL_SERIALIZE( pending);
+            )
+         };
```
lazan, line 46: Simple and clean.

middleware/file/include/file/manager/handle.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace casual
+{
+   namespace file::manager::handle
+   {
+
+      using dispatch_type = decltype( common::message::dispatch::handler( common::communication::ipc::inbound::device()));
+
+      //! @returns all the handlers for service manager
+      dispatch_type create( State& state);
```
lazan, line 22: _file-manager_

middleware/file/source/api/c/file.cpp
```diff
@@ -0,0 +1,129 @@
...
+         {
+            namespace global
+            {
+               thread_local std::error_code code;
+            } // global
+         } //
+      } // local
+
+
+      extern "C"
```
lazan: I don't think you need `extern C` in the TU

middleware/file/source/manager/resource.cpp
lazan, line 1: I strongly prefer not to send anything from _resouce.cpp_. If possible all send should be in _handle.cpp_ . It's more symmetric.  receive-request -> send-reply at in the same function. 
This would mean that you return the message that _handle_ should send back to caller. Then it's easy to unittest the resource as a bonus.

middleware/file/source/manager/handle.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+               namespace commit
+               {
+                  auto request( State& state)
+                  {
+                     return [ &state]( const common::message::transaction::resource::commit::Request& message)
+                     {
+                        common::Trace trace{ "file::manager::handle::local::transaction::commit::request"};
+
+                        resource::commit( state, message);
```
lazan, line 80: I would like `resource::commit(..)` to return the message and this function sends the reply to caller. Much clearer that the reply is sent.

lazan (changes requested):
First pass :)

middleware/file/documentation/file.development.md
KristianIvarsson: Good idea, we'll put that into some commit sometime

middleware/file/include/file/api/c/file.h
KristianIvarsson: The documentation (description) is wrong though

middleware/file/include/file/factory.h
KristianIvarsson: Hmm, it must be some leftovers to avoid ODR in some earlier design and I guess it can be removed now

middleware/file/source/manager/resource.cpp
KristianIvarsson: Maybe agreed, but replies are not always gonna be sent but instead sent in other scenarios, such as when a _commit_ is issued, maybe replies for _reserve_ is gonna be sent (as well)

Maybe I don't understand what you're aiming for ?

middleware/file/include/file/factory.h
lazan, line 1: This is a _message factory_? 
* should not the file name reflect that?
* why header only?

In the rest of `causal` I think _factory_ is used for dynamic dispatch factories (sort of classic factory). We have very few. 
This is more of a _create helper_? 
`file/message/create.h` ?

middleware/file/source/manager/handle.cpp
```diff
@@ -0,0 +1,143 @@
...
+               {
+                  auto request( State& state)
+                  {
+                     return [ &state]( const common::message::transaction::resource::prepare::Request& message)
+                     {
+                        common::Trace trace{ "file::manager::handle::local::transaction::prepare::request"};
+                        
+                        auto reply = common::message::reverse::type( message);
+                        state.multiplex.send( message.process.ipc, reply);
+                     };
```
lazan, line 68: Have a `resource::prepare` for symmetry.

middleware/file/include/file/factory.h
KristianIvarsson: Give a name suggestion

It is used from 2 binaries and thus we need to build it as a object file or an archve and link it into those two binaries so this was just simpler to have it as header only

lazan: I just gave a name suggestion ..?
Yeah, either an archive or you just keep the object-file(s) in a python-list and add them in the binaries that needs them.

middleware/file/source/manager/handle.cpp
KristianIvarsson: Greate point (fixed now)

middleware/file/include/file/factory.h
lazan: `file::message::create::[non::]blocking::request( ...)`

KristianIvarsson: Did some more proper renaming of namespaces (and removed a forgotten ìnline` directive, but didn't change any file structure)

lazan (approved):
Nice.

PR #496, opened 2025-01-13
Patch/1.7/simplified stale

lazan:


middleware/transaction/source/manager/handle.cpp
```diff
@@ -439,6 +381,45 @@ namespace casual
...
+                     // There should almost always be a a caller to the rollback/commit, but we might have started a rollback our selves.
+                     if( destination.process != common::process::handle())
+                     {
+                        common::log::line( verbose::log, "reply: ", reply);
+                        state.multiplex.send( destination.process.ipc, reply);
+                     }
+                     else
+                     {
+                        // We only start rollbacks our selves.
+                        CASUAL_ASSERT( ( concepts::any_of< Reply, common::message::transaction::rollback::Reply>));
```
AlbsAgren, line 398: Why both the assert and constexpr if?

AlbsAgren (approved):
Looks good!

middleware/transaction/source/manager/handle.cpp
lazan: The `constexpr if` is to only generate code for rollback-reply. Otherwise we get a compilation error, because it could be a commit-reply, even though we should never get to this branch with a commit-reply. Which the `assert` validates.

lazan: Maybe it would be better if we split and overload the function for the two reply-types?

PR #500, opened 2025-01-14
Patch/1.7/merge 1.6.23

AlbsAgren:


middleware/service/source/manager/handle.cpp
```diff
@@ -69,16 +69,28 @@ namespace casual
...
+
+                  if( state.events.active< common::message::event::service::Calls>())
+                  {
+                     common::message::event::service::Metric metric;
+                     metric.code.result = code;
+                     metric.correlation = caller.correlation;
+                     metric.service = caller.service;
+                     metric.process = callee;
+
+                     state.metric.add( metric);
```
lazan: `std::move( metric)`

middleware/service/include/service/manager/state.h
```diff
@@ -97,6 +97,7 @@ namespace casual
                common::process::Handle process;
                common::strong::correlation::id correlation;
                common::transaction::global::ID gtrid;
+               std::string service;
```
lazan: use `state::service::id::type` instead?

middleware/service/source/manager/handle.cpp
```diff
@@ -69,16 +69,28 @@ namespace casual
...
 
                   state.multiplex.send( caller.process.ipc, std::move( message));
+
+                  if( state.events.active< common::message::event::service::Calls>())
+                  {
+                     common::message::event::service::Metric metric;
+                     metric.code.result = code;
+                     metric.correlation = caller.correlation;
+                     metric.service = caller.service;
+                     metric.process = callee;
```
lazan: I want us to provide the transaction id, if we got it.
In `Caller` we got `gtrid`. We could just create an transaction::ID from `gtrid` with no `bqual`, or keep a `transaction::ID` in `Caller` instead of `transaction::global::ID`

middleware/service/source/manager/handle.cpp
```diff
@@ -278,7 +290,7 @@ namespace casual
                            log::error( code::casual::invalid_semantics, " callee terminated with pending reply to caller - callee: ", 
                                  event.state.pid, " - caller: ", caller.process.pid);
 
-                           local::error::reply( state, caller, common::code::xatmi::service_error);
+                           local::error::reply( state, caller, common::code::xatmi::service_error, common::process::Handle{ event.state.pid});
```
lazan: I think `state.remove(...)` should return what's needed to send error reply.
A vector of _removed-wrapper_. that holds `Caller` and `process::Handle` (callee). If anybody knows this, it's `state.remove( ...)`, after the call, the related state to the pid is gone.

middleware/service/source/manager/handle.cpp
```diff
@@ -574,7 +586,7 @@ namespace casual
                            {
                               if( message.no_reply())
                                  return {};
-                              return { .process = message.process, .correlation = message.correlation, .gtrid = message.gtrid};
+                              return { .process = message.process, .correlation = message.correlation, .gtrid = message.gtrid, .service = message.requested};
```
lazan: use  `.service = service_id`

lazan (changes requested):

middleware/service/include/service/manager/state.h
lazan: And remove `m_reserved_service` and use `caller().service` to get _reserved-service_. i think this will be an improvement.

AlbsAgren: Had a go at it, maybe a little awkward on TPNOREPLY where caller is empty? Have a look.

middleware/service/include/service/manager/state.h
```diff
@@ -96,15 +96,28 @@ namespace casual
...
                CASUAL_LOG_SERIALIZE(
                   CASUAL_SERIALIZE( process);
                   CASUAL_SERIALIZE( correlation);
-                  CASUAL_SERIALIZE( gtrid);
+                  CASUAL_SERIALIZE( trid);
+                  CASUAL_SERIALIZE( service);
+               )
+            };
+
+            struct Reservation
```
lazan, line 113: Good name.

lazan (approved):
Super!

lazan:
Merged manually

PR #502, opened 2025-01-18
log: simplified logging - removed all verbose

lazan:
This patch removes all "manager" specific trace and verbose logging. If one only wants trace for a specific "manager", only turn on logging/trace for that "manager".


KristianIvarsson (approved):
Looking good

PR #504, opened 2025-01-21
common: fix common::file::permission::execution

AlbsAgren:
Before: we didnt check if the path existed before checking its premission, causing the function to always return true for non-existent paths.
Now: we return false if the path does not exist.

### DIFF
--- a/middleware/common/source/file.cpp
+++ b/middleware/common/source/file.cpp
@@ -194,7 +194,11 @@ namespace casual
             bool execution( const std::filesystem::path& path)
             {
                namespace fs = std::filesystem;
-               return ( fs::status( path).permissions() & fs::perms::owner_exec) != fs::perms::none;
+               auto status = fs::status( path);
+               if( status == fs::file_status( fs::file_type::not_found))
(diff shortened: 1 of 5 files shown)


middleware/common/source/file.cpp
```diff
@@ -194,7 +194,11 @@ namespace casual
             bool execution( const std::filesystem::path& path)
             {
                namespace fs = std::filesystem;
-               return ( fs::status( path).permissions() & fs::perms::owner_exec) != fs::perms::none;
+               auto status = fs::status( path);
+               if( status == fs::file_status( fs::file_type::not_found))
+                  return false;
+
+               return ( fs::status( path).permissions() & fs::perms::owner_exec) == fs::perms::owner_exec;
```
lazan: Why not `return ( status.permissions() & ... )`

AlbsAgren: Good question

lazan (approved):
Nice. Very strange that we've not notice this before...

lazan (approved):

PR #513, opened 2025-01-30
common: use std::format for time-points based on __cpp_lib_format_path

### LINKED ISSUE #512: use std::format for time-points based on __cpp_lib_chrono macro
https://en.cppreference.com/w/cpp/chrono

```
201907L	(C++20)	Calendars and Time zones
```

lazan:
Before: we used the macro `__APPLE__` to treat only _apple_ different

Now: we use the test macro `__cpp_lib_format_path` to decide if we're using std::format for std::chrono::time_point.

Resolves #512

We only merge this if we get som more fixes for 1.7.

### DIFF
--- a/middleware/common/source/chronology.cpp
+++ b/middleware/common/source/chronology.cpp
@@ -16,13 +16,29 @@
 #include <functional>
 #include <algorithm>
 
+
 namespace casual
 {
    namespace common::chronology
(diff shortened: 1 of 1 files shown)


lazan:
I was way off!!! we're did I get path from??

PR #523, opened 2025-02-13
common log: Activate log used in unittest

### LINKED ISSUE #503: unnittest common_log_stream, relocate__expect_content_in_new_location fails (when logging is off)
On MacOS

OlofGullnas:
A unit test logged to debug, but debug logging was not active. This resulted in not finding the logged data in the log and a failed unit test.
Before change "casual.common.verbose" was activated, now "common.debug" is activated. At one time the test logged to verbose but that was changed to debug some time ago.
 
Fixes issue #503

### DIFF
--- a/middleware/common/unittest/source/log/test_stream.cpp
+++ b/middleware/common/unittest/source/log/test_stream.cpp
@@ -86,7 +86,7 @@ namespace casual
 
          auto scope = local::log::scoped::path( path);
 
-         log::stream::activate( "casual.common.verbose");
+         log::stream::activate( "casual.debug");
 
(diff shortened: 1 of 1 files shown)


lazan:
You need to update your commit message. We follow "standard" formatting that `vim` provides.

```
<module>: <subject in lower case, the total length must not exceed 70 characters>
<empty line>
<description of what is changed, every line must not exceed 70 characters>
...

<possible references to issues>
```

Which would render something like:

```
 common: activate log used in unittest
 
 A unit test logged to debug, but debug logging was not active. This resulted
 in not finding the logged data in the log and a failed unit test.
 
 Fixes issue #503
 ```

lazan (approved):
Looks good.

PR #524, opened 2025-02-13
Feature/1.8/http inbound improve

AlbsAgren:
Since the unittests are dependent on the inbound having been built, it proved easier to integrate the whole thing into the main build. It's not very pretty unfortunately.

The tests run fine on the ubi8-builder, save for a curl option that isnt supported. They dont run on the SUSE one, but I'm publishing this draft for input before I look into it further.


middleware/http/source/inbound/nginx/plugin/makefile.cmk
```diff
@@ -1,22 +1,89 @@
...
-   print(out.decode())
+def perform_clean(dummy):
+   check_preconditions()
+   if nginx_makefile_exists():
+      print('nginx - running clean')
+      print( subprocess.check_output(['make', 'clean'], cwd = NGINX_SOURCE_DIR).decode())
 
 # main - register function
-model.register( 'nginx').execute(True).serial(True).add_recipe( Recipe( perform_build, None))
+inbound_build_target = model.register('nginx-build').execute(True).serial(True).add_recipe( Recipe( perform_build, None))
```
lazan, line 81: Why is this _serial_?

lazan (approved):
I find it pretty easy to understand.

But, if you do a regular _build_, `nginx` will not be build? But if you do a `cmk test` nginx will be _built_ since the unittest has dependency?

middleware/http/source/inbound/nginx/plugin/makefile.cmk
AlbsAgren: Good point. Thinking about it we could also `make -q` to test if it needs executing at all, but casual-make doesnt support passing `False` to `execute` right now.

AlbsAgren:
Yes, thats correct. That feels inconsistent, I'll see if I can make it better.

lazan:
> Yes, thats correct. That feels inconsistent, I'll see if I can make it better.

I think if _nginx-build_ is more a regular _build_, and treated as everything else, it would work. Keep the _nginx-install_.

lazan:
I was thinking, maybe we should use some other port than `80` or `8080`.  It's pretty high probability that users already using these. We should use something in the regular unittest range `[ 7000, 8000)`.

middleware/http/source/inbound/nginx/plugin/makefile.cmk
AlbsAgren: It suffers from the same issue with pickle as the unittests. The short version is that its impossible to use functions in cmk files with _concurrent_. It might be possible to fix casual-make but its probably not a quick fix.

test/unittest/source/test_http.cpp
```diff
@@ -116,12 +117,29 @@ http {
...
+                  {
+                     auto capture = administration::unittest::cli::command::execute( curl);
+                     if( capture)
+                        return;
+                     
+                     common::process::sleep( std::chrono::milliseconds{ 10});
+                  }
+
+                  common::code::raise::error( code::casual::invalid_semantics, "failed to get a response from inbound after 1000 attempts");
+               }
```
lazan, line 127: Use `unittest::eventually::succeed` instead, and just do a lambda that returns false until it succeed.

PR #528, opened 2025-03-05
file: adding cli for administrating casual-file

KristianIvarsson:
Contains mechanisms to list reserved paths and recover transactions

This is the first attempt for the first version of CLI to casual-file

It just contains two mechanisms so far, i.e. 

- List ongoing reserved requests
- Recover transactions (by commit or rollback)


middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,174 @@
...
+                        formatter.print( std::cout, state.requests);
+                     };
+
+                     return argument::Option{
+                        std::move( invoke),
+                        { "-r", "--list-reservations"},
+                        "list information of files currently reserved"
+                     };
+                  }
+               } // files
```
lazan: reservations

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,174 @@
...
+                     }
+
+                     auto complete()
+                     {
+                        return []( bool help, auto values) -> std::vector< std::string>
+                        {
+                           if( help)
+                              return { "<gtrid>"};
+
+                           return { "<value>"};
```
lazan: You could call state, and return all gtrid that the manager knows about. user can then tap-complete gtrids

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+      argument::Option options()
+      {
+         return argument::Option
+         { [](){}, { "file"}, "file related administration"}
+         ({
+            local::list::reservations::option(),
+            local::assets::recovery::commit::option(),
+            local::assets::recovery::rollback::option(),
+         });
```
lazan, line 176: To be "conformant", we should provide a `--state`.

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,174 @@
...
+               namespace format
+               {
+                  auto requests( const manager::admin::model::State& state)
+                  {
+                     const auto color = common::terminal::output::directive().porcelain() ? 
+                        common::terminal::color::yellow : 
+                        common::terminal::color::no_color;
+
+                     return common::terminal::format::formatter< manager::admin::model::Request>::construct(
+                        common::terminal::format::column( "process", []( const auto& r){ return r.process;}, color),
```
lazan: We don't use `process` anywhere else. Should be `pid`.

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,174 @@
...
+               } // call
+
+               namespace format
+               {
+                  auto requests( const manager::admin::model::State& state)
+                  {
+                     const auto color = common::terminal::output::directive().porcelain() ? 
+                        common::terminal::color::yellow : 
+                        common::terminal::color::no_color;
+
```
lazan: You don't need to check porcelain at all. Use the colors you want, the colors will be ignored if user uses `--porcelain`

In a few places we check for _porcelain_ to keep backward compatibility -> the order of stuff is the same as before when using _porcelain_. You don't have to bother with this since it's all new stuff.

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,174 @@
...
+                     return common::terminal::format::formatter< manager::admin::model::Request>::construct(
+                        common::terminal::format::column( "process", []( const auto& r){ return r.process;}, color),
+                        common::terminal::format::column( "gtrid", []( const auto& r){ return r.gtrid;}, color),
+                        common::terminal::format::column( "stage", []( const auto& r){ return description( r.stage);}, color),
+                        common::terminal::format::column( "path", []( const auto& r){ return r.path;}, color));
+                  }
```
lazan, line 64: The "pattern" we've used for other stuff: 
`<main-stuff yellow>  <other related stuff> ... <timestamp or other time stuff blue>` 
Hence, it should be something like:
`<path yellow> <gtrid not yellow>> <stage not yellow>  <pid white> <timestamp blue>`  

I think `timestamp` would be a useful addition for the user.

middleware/file/source/manager/admin/cli.cpp
KristianIvarsson: What should it consist of ?

lazan: Whatever the _state service_ replies with. Should be _admin Model_. 
Just add:
```cpp
         local::assets::recovery::commit::option(),
          local::assets::recovery::rollback::option(),
             casual::cli::state::option( &local::some::function::to:call::state::and::return::the::model),
         });
 ```
 ... and the ` casual::cli::state::option` will take care of the `--state` option and print it for the user and so on.

File: middleware/file/source/manager/admin/cli.cpp (hunk 6 above)
Comment (KristianIvarsson) [reply middleware/file/source/manager/admin/cli.cpp:64]: `timestamp` is now added

File: middleware/file/source/manager/resource.cpp [hunk 7]
```diff
@@ -101,66 +101,104 @@ namespace casual::file::resource
...
             {
                rollback( request);
             }
-         } // detail
 
+            auto reserve( State& state, const Request& request) -> std::expected<std::filesystem::path, code>
+            {
+               if( state.runlevel == State::Runlevel::shutdown)
+               {
+                  common::log::error( common::exception::capture(), "failed to reserve due to shutdown");
```
lazan: `common::exception::capture()` expects to capture an exception in-flight. Is there an exception? If so, I find it error-prone to use `common::exception::capture()` outside a _catch block_.

middleware/file/source/manager/resource.cpp
```diff
@@ -101,66 +101,104 @@ namespace casual::file::resource
...
+                  // requested path is involved
+   
+                  if( work->trid == request.trid)
+                  {
+                     return temporary( request.trid, request.path);
+                  }
+                  else
+                  {
+                     return std::unexpected( code::busy);
+                  }
```
lazan: [nitpick] Why not just
```cpp
               if( work->trid == request.trid)    
                     return temporary( request.trid, request.path);
                  else
                     return std::unexpected( code::busy);
```
?

middleware/file/source/manager/resource.cpp
```diff
@@ -101,66 +101,104 @@ namespace casual::file::resource
...
+            }
+   
+         } // detail
 
-         auto reserve( State& state, const Reserve& request) -> std::expected<std::filesystem::path, code>
+         void reserve( State& state, Request request)
          {
-            if( state.runlevel == State::Runlevel::shutdown)
+            auto reply = common::message::reverse::type( static_cast<const Reserve&>(request));
+
```
lazan, line 147: Hmm... Maybe it's better to specialize 
```cpp
   namespace common::message::reverse
   {
      template<>
      struct type_traits< your-sub-type, reverse-type>
    }
    ```
    and get rid of the `static_cast`?

File: middleware/file/source/manager/resource.cpp [hunk 10]
```diff
@@ -101,66 +101,104 @@ namespace casual::file::resource
...
-
-            if( const auto work = std::ranges::find( state.working, request.path, &Reserve::path); work != state.working.end())
+            else
             {
-               //
-               // requested path is involved
-
-               if( work->trid == request.trid)
+               reply.code = std::move( result.error());
+            }
```
lazan: remove the scope and just one-liners?

middleware/file/source/manager/resource.cpp
```diff
@@ -187,66 +225,28 @@ namespace casual::file::resource
...
          }
-
       } // 
    } // local
 
-   void reserve( State& state, Reserve request)
+   void reserve( State& state, const Reserve& request)
    {
-      request.path = request.path.lexically_normal();
+      //common::log::error( common::exception::capture(), "working ", state.working.size());
```
lazan: remove

middleware/file/source/manager/resource.cpp
```diff
@@ -101,66 +101,104 @@ namespace casual::file::resource
...
+               if( state.runlevel == State::Runlevel::shutdown)
+               {
+                  common::log::error( common::exception::capture(), "failed to reserve due to shutdown");
+                  return std::unexpected( code::error);
+               }
+   
+               if( const auto work = std::ranges::find( state.working, request.path, &Request::path); work != state.working.end())
+               {
+                  //
+                  // requested path is involved
```
lazan, line 112: Why the empty `//`?

middleware/file/source/manager/resource.cpp
```diff
@@ -187,66 +225,28 @@ namespace casual::file::resource
 
                //
                // free related (potentially) pending requests for this path (for real)
-               for(auto&& wait : pull( state.pending, work.path, &Reserve::path))
+               for(auto&& wait : pull( state.pending, work.path, &Request::path))
```
lazan, line 199: Why `auto&&`? `pull(..)` either return by value or reference, right?

middleware/file/source/manager/admin/server.cpp
```diff
@@ -33,14 +33,15 @@ namespace casual
...
+                     return [stage=stage]( const manager::State::Request& value)
                      {
                         return admin::model::Request
                         {
                            .pid = value.process.pid,
                            .gtrid = common::transaction::id::range::global( value.trid),
-                           .path = value.path,
                            .stage = stage,
+                           .path = value.path,
+                           .time = value.time,
```
lazan, line 44: I find this so easy to read!

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -59,6 +59,7 @@ namespace casual
                         common::terminal::format::column( "pid", []( const auto& r){ return r.pid;}, common::terminal::color::white),
                         common::terminal::format::column( "gtrid", []( const auto& r){ return r.gtrid;}),
                         common::terminal::format::column( "stage", []( const auto& r){ return description( r.stage);}),
+                        common::terminal::format::column( "time", []( const auto& r){ return std::chrono::floor< std::chrono::microseconds>( r.time);}, common::terminal::color::blue),
                         common::terminal::format::column( "path", []( const auto& r){ return r.path;}, common::terminal::color::yellow));
                   }
```
lazan, line 64: Why do you want the most significant last? The users "key" is the _path_, it should be the first column. It's consistent with everything else, and user can easy use `sort` and stuff.
Please make it the same look'n feel, pretty please with sugar on top 😄

middleware/file/source/manager/admin/cli.cpp
```diff
@@ -0,0 +1,179 @@
...
+                           if( help)
+                              return { "<gtrid>"};
+
+                           std::vector< std::string> result;
+
+                           std::ranges::transform( 
+                              local::detail::call::state().requests,
+                              std::back_inserter( result),
+                              []( const auto& request) { return common::string::compose( request.gtrid); });
+
```
lazan, line 128: Why not use `algorithm::transform(...)` that returns an std::vector with the return type of the functor? Hence, you can do a "one-liner".
The most common use-case for transform in `casual` is exactly this -> we have our own algorithm for it.

middleware/administration/unittest/source/cli/test_file.cpp
```diff
@@ -0,0 +1,126 @@
...
+         auto domain = local::domain();
+
+         const auto file = common::unittest::file::temporary::name( ".txt");
+
+         ASSERT_EQ( common::transaction::context().begin(), common::code::tx::ok);
+
+         const auto reserved = file::blocking::reserve( file);
+
+         const auto capture = local::execute( "casual file --list-reservations --porcelain true").standard.out;
+
```
lazan, line 88: I've started to have a comment the shows the expected output
https://github.com/casualcore/casual/blob/feature/1.8/main/middleware/administration/unittest/source/cli/test_service.cpp#L77

I think this is very helpful to understand the tests.

To get the output I just do a temporary 
```cpp
auto capture  = some_cli_stuff;
EXPECT_TRUE( false) << capture.standard.out;
```

middleware/administration/documentation/cli/file.operation.md
lazan, line 1: Perfect, you did remember to actually generate the documentation 😄 We've missed this to many times.

middleware/file/source/manager/resource.cpp
KristianIvarsson: Brackets are there for some kind of consistency, hence there are a few if-statements with just one statement with brackets

middleware/file/source/manager/resource.cpp
KristianIvarsson: Oups, will do

middleware/file/source/manager/resource.cpp
KristianIvarsson: That is of course an option. How strong do the reviewer feel about it?

middleware/file/source/manager/admin/cli.cpp
KristianIvarsson: Well, path's are of different lengts and thus the _table_ will jump around, i.e. the content with variable length is the last one that makes the previous _columns_ to be "stable"

middleware/file/source/manager/admin/cli.cpp
KristianIvarsson: Well, I didn't bring in anything of the `casual algorithm` functions at all and thought this one thing (that soon can be replaced by `std::ranges::to<>`) wasn't a reason enough to break the std-look-and-feel

middleware/administration/documentation/cli/file.operation.md
KristianIvarsson: It was tricky though

middleware/administration/unittest/source/cli/test_file.cpp
KristianIvarsson: Fixed

Good idea

middleware/file/source/manager/resource.cpp
KristianIvarsson: Well spotted

It has to be some IDE thing and a sloppy programmer or just a copy/paste thing followed by a sloppy programmer

middleware/file/source/manager/admin/cli.cpp
KristianIvarsson: Fixed

middleware/file/source/manager/resource.cpp
KristianIvarsson: Fixed

middleware/file/source/manager/resource.cpp
KristianIvarsson: "Fixed"

middleware/file/source/manager/resource.cpp
KristianIvarsson: Fixed

middleware/file/source/manager/admin/cli.cpp
KristianIvarsson: Fixed (being grumpy)

middleware/file/source/manager/admin/cli.cpp
lazan: But why not use `algorithm`?  I don't understand. Of course we have common use cases that "custom" algorithms will make it easier.

KristianIvarsson: I already motivated my answer and in this case I rather change it when we get full C++23 support, because current implementation is not wrong

lazan (approved):

PR #529, opened 2025-03-06
publiccode.yml

cklellie:
In order to publish to the European Open Source Software Catalogue we need to be registered with offentligkod.se and have a `publiccode.yml` file.

An initial version of the file has been created, it is a work in progress with this PR serving as a mechanism to review and update the contents of the `publiccode.yml` appropriately before it is merged.

The specification for publiccode.yml is available here: 
https://yml.publiccode.tools/index.html
https://github.com/publiccodeyml/publiccode.yml

There are also github actions and validators to assist with lifecycle management of this file.

Prior to the merge this PR the file should be check with a validation tool to ensure it is well formed.

### DIFF
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -1,0 +1,79 @@
+publiccodeYmlVersion: "0.4"
+
+name: Casual
+applicationSuite: #optional
+url: "https://github.com/casualcore/casual.git"
+landingUrl: "https://casual-software.com/" #optional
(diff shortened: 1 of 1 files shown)


lazan:
@cklellie Change your branch name to `feature/1.8/publiccode-documentation`. We use the branch name to deduce what tooling to use when building.

PR #548, opened 2025-04-17
Patch/1.7/http trace span

lazan:


middleware/http/source/inbound/call.cpp
```diff
@@ -47,10 +47,31 @@ namespace casual
...
+               {
+                  if( auto found = algorithm::find( header, http::header::name::execution::trace::parent))
+                  {
+                     auto field = algorithm::container::extract( header, std::begin( found));
+
+                     return detail::transform::span( field.value());
+                  }
+                  else
+                  {
+                     common::log::error( code::bad_request, "missing traceparent header - action: ignore");
```
partickl: should this really be considered an error? imo we shouldn't require tracing info from caller

middleware/http/source/inbound/call.cpp
```diff
@@ -216,6 +241,30 @@ namespace casual
...
       } // local 
 
+      namespace detail
+      {
+         namespace transform
+         {
+
+            auto span( std::string_view value) -> std::tuple< common::strong::execution::id, common::strong::execution::span::id>
+            {
+               static const std::regex regex{ R"(^00-[0-9a-f]{32}-[0-9a-f]{16}-[0-9][0-9]$)"};
```
partickl: last 2 characters are specified as hex in the specification, but at the same time constrained to only 00 or 01 in this version, so not a big issue

partickl (approved):

middleware/http/source/inbound/call.cpp
lazan: I concur.

partickl (approved):

PR #563, opened 2025-05-26
Patch/1.7/build rm

### LINKED ISSUE #561: casual-build-resource-proxy gets compile error
I get the following error when compiling a resource proxy:
```
execute g++ - failed
build resource proxy - failed
[casual:invalid_argument] failed to build proxy - capture: { exit: 1, standard: { out: , error: rm_proxy_3ecf11756d3643b7a83bbb069f370f64.cpp: In function ‘int main(int, const char**)’:
rm_proxy_3ecf11756d3643b7a83bbb069f370f64.cpp:28:10: error: invalid conversion from ‘const char**’ to ‘char**’ [-fpermissive]
   28 |          argv,
      |          ^~~~
      |          |
      |          const char**
}}: invalid_argument

```
I am using gcc-13. I suspect this worked ok with an older compiler..

If I add a compiler directive "-fpermissive" the build goes OK.

Seems as if this is caused by the casual-build-resource-proxy template for the generated code declaring main as: 

`int main( int argc, const char** argv)
`
The error occurs when const char** argv is used in the initalization of serverArguments.

The build goes OK if "-fpermissive" is added with  "build-directives". This should not be needed. Removing the "const" from the argv argument to main also "works".

(I discovered this whien I started working on an "XA-driver" for PostgreSQL. Postgres does not support XA out of the box, Glue code is required. Using iPostgres with Casual XA will have limitations/restrictions, IF I can get it to work in a reasonable way... I have not built a resource proxy in a long time, but now I needed to build one. My work is still in an early stage. I only have a skeleton for the "PosgresSQL XA driver".)

### LINKED ISSUE #562: casual-build-resource-proxy --keep-sources does not work when build fails
When the compiile of the generated file fails (as in issue #561) the generated source file is not kept even if "--keep-sources 1" is specified.
If the buikd goes OK the generated file is kept if requested.

As one use case for --keep-sources is to investigate failures to build the generated code this is a bit inconvenient. This MAY BE the intended behaviour, but it is not what I expected. My first reaction when I got the error was to add --keep-sources to check what the file looked like.

I suspect this is caused by an exception generated when the build fails, and that this bypasses the path.release() call that stops the delete/unlink of the   file (line 241 in tools/source/build/resource/proxy/main.cpp) . 

Can be fixed by enhancments to common::file::scoped::Path to provide a mechanism to indicate that the delete should not be done (a "flag" can be set when  setting up the object) or by catching the exception... And possibly other ways.

I do not know if the same issue exist in the other casual build tools...

lazan:


middleware/tools/source/build/resource/proxy/main.cpp
```diff
@@ -227,6 +228,13 @@ int main( int argc, const char** argv)
 
                         // Generate file
                         common::file::scoped::Path path( common::file::name::unique( "rm_proxy_", ".cpp"));
+                        
+                        // make sure we keep the file if user has requested it
+                        auto path_keep = common::execute::scope( [keep = settings.keep_source, &path]()
+                        { 
+                           if( keep)
+                              path.release();
```
OlofGullnas, line 236: Is the code at lines 249-252 still needed? As I undertstand this I suspect not. I neded to think a few seconds before I undrstood how this worked. It obviously relies on path_keep executing its lambda before path is destructed. Should be the case as objects are destroyes in reverse order of construction...

lazan:
```cpp
                        if( settings.keep_source)
                        {
                           path.release();
                        }
 ```
 
 yeah, this is not needed any more.

lazan: > It obviously relies on path_keep executing its lambda before path is destructed. Should be the case as objects are destroyes in reverse order of construction...

Exactly, and since we need to take the thing we want to alter in some way via the _guard_, the lifetime sort it self out automatically.

ahhud (approved):

OlofGullnas:
@lazan Noticed you have assigned me to the pull request. Intentional?

A "Rebase and merge" button is now enabled, and no review button,

Do I have merge rights?

I do approve the PR.

One small comment.The code still supports one-letter options to the command, but they are removed from the proxy.md description. I can understand them being "deprecated"...

The single letter variants of options are still described in for example "casual domain --help".

I also nioticed that if only --help is specified the command complains about "invalid argument". Should a command with --help be treated as a request for help information without an actual build attempt? The current behaviour is OK but perhaps a little surprising.
What is customary in Unix applications/commands with a --help option?  E.g. cc does not complain about missing input on a "cc --help" command, but I think I have seen others that behave like "casual-build-resource-proxy --help". Changing the behaviour  would obviously change behaviour of existing command scripts with --help and a "comlete set" of build options...

OlofGullnas (approved):

lazan:
> The single letter variants of options are still described in for example "casual domain --help".

Yeah... Thats the "view" from the top argument tree. This comes from when you do `casual --help` you probably don't want 4 pages of help for every suboptions and so on. So this is kind of a hack to make that better. 
And this of course is the consequence of that hack. 
Should be fixed in som way, maybe that we only do "one level" per help sort of. It would be nice if you write an issue for this.

lazan:
> Do I have merge rights?
> 
> I do approve the PR.

I'll merge it manually.

lazan:
> What is customary in Unix applications/commands with a --help option?

This should work. `casual-build-resource-proxy --help --help`, for example

PR #565, opened 2025-06-06
Feature/1.8/new casual make model

ahhud:
This PR has a dependency to https://github.com/casualcore/casual-make/pull/10.
In order to be able to build the .github/workflows/casual-build-alpha.yaml temporarly reference a different branch.
This needs to be corrected before merge.

### DIFF
--- a/.casual-make/configuration.json
+++ b/.casual-make/configuration.json
@@ -1,0 +1,448 @@
+{ 
+  "configuration" : 
+  {
+    "version" : 1.0,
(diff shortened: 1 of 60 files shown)


lazan (approved):
I bit asymmetric with include and lib-path, but I guess this is the only way.

lazan:
@ahhud I think you should rebase and build locally before you merge. It's likely that 1.8/main has makefiles that has changed.

PR #574, opened 2025-07-05
Feature/1.8/kqueue epoll

lazan:


middleware/common/include/common/communication/select/epoll.h
```diff
@@ -0,0 +1,163 @@
...
+#include <vector>
+
+namespace casual
+{
+   namespace common::communication::select
+   {
+      namespace directive
+      {
+ 
+         using range_type = range::type_t< std::vector< strong::file::descriptor::id>>;
```
KristianIvarsson: `type_t` has to be a tautologi because `t` is an abbreviation for `type`

middleware/common/include/common/communication/select/epoll.h
```diff
@@ -0,0 +1,163 @@
...
+            template< typename Range>
+            auto add( Range&& descriptors) noexcept
+               -> decltype( add( range::front( descriptors)))
+            {
+               algorithm::for_each( descriptors, [&]( auto descriptor){ add( descriptor);});
+            }
+
+            template< typename Range>
+            auto remove( Range&& descriptors) noexcept
+               -> decltype( remove( range::front( descriptors)))
```
KristianIvarsson: There's no return  so having trailing return type declaration seems to be ambigious

lazan: I'm using trailing return type with _expresion template_ (i think it's called) to make sure that elements in the `Range` can be used. If not `decltype( remove( range::front( descriptors)))` evaluates to a type (void in this case) the whole function is not instantiated.
Like `enable_if`, but I think this _technique_ is much cleaner and easier to read.

KristianIvarsson: Apparently it is not easier to read if the purpose was misunderstood ;-)

middleware/common/include/common/communication/select/epoll.h
lazan: Yeah... I tried to mimic `std` _helper types_.  Such as `std::remove_reference` has a _helper type_ `std::remove_reference_t`

middleware/common/include/common/communication/select/epoll.h
lazan: I think you'll prefer this over enable_if boilerplate. 
But, now we could use ad-hoc `requires` which might be more informative.

lazan: Apparently I updated it to concepts.

PR #591, opened 2025-08-11
tools: normalized options for build-server/executable

### LINKED ISSUE #567: casual-build-executable missing --system-configuration option?
casual-build-executable does not seem to have a --system-configuration option. This option is present in for example casual-build-resource-proxy.

There seems to be some code in tools/source/build/executable/main.cpp in support for this option. "settings.files.system" is tested in transform::state(const Settings& settings) but is not set when parsing the command line.

Might be a missing option when parsing command line in main().

I could work around this by using the deprecated CASUAL_RESOURCE_CONFIGURATION_FILE environment variable.

(Or by adding a line defining --system-configuration to the parse() call, copied from options for building resource proxy).

I also noticed that options for specifying "keep sources", and additional compiler and linker directives differs between the build commands for resource proxies, executables and servers.

Conformity would be good. Easier to remember options. Changing things may affect backward compatibility. Adding (optional) options is usually OK

I also think the casual-build-server command has an option for specifying resource keys on the command line. No such option when building executable. Should such a mechanism also exist when building executables? Useful for anything?  I specify resources in the definition file for executable and server. This seems more natural to me, and also allows specifying resource names.

lazan:
This patch: Normalizes common options that both build-server and build-executable uses.
This will help keeping the two from diverging.

Resolves #567


OlofGullnas:
I have started looking at this PR.

Factoring out the common part of the command options for the "build tools" is a good approach :-)

I tried to build the branch, but I get an error:

```
buildserver: /export/home/guo/c_1.8/casual-1.8/middleware/example/server/bin/casual-example-server
buildserver: /export/home/guo/c_1.8/casual-1.8/middleware/example/server/bin/casual-example-resource-server
error:
[casual:invalid_argument] failed to compile - capture: { exit: 1, standard: { out: , error: /usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /export/home/guo/c_1.8/casual-1.8/middleware/common/bin/libcasual-mockup-rm.so: undefined reference to `casual::common::transaction::xid::null()'
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /export/home/guo/c_1.8/casual-1.8/middleware/common/bin/libcasual-mockup-rm.so: undefined reference to `casual::common::transaction::Context::instance()'
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /export/home/guo/c_1.8/casual-1.8/middleware/common/bin/libcasual-mockup-rm.so: undefined reference to `casual::common::unittest::log'
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /export/home/guo/c_1.8/casual-1.8/middleware/common/bin/libcasual-mockup-rm.so: undefined reference to `casual::common::transaction::Context::resource_registration(casual::common::strong::Type<int, casual::common::strong::resource::policy>, xid_t*)'
collect2: error: ld returned 1 exit status
}}: invalid_argument

aborting due to errors
error building...
[infra] guo@bvin01-unisb-01:~/c_1.8/casual-1.8 
``` 
The build error is not obviously cobbected to the changes in the PR.

I will start a new build after a "casual-make --debug clean" in case this is caused by a inconsistent build. 

(Or is this caused by the build referencing an old "installed" verion of casual. or something like that. My last build of Casual was some time ago,)

OlofGullnas:
There was an old file common/bin/libcasual-mockup-rm.so in the build tree, When I removed it the build worked OK....

Will continue looking into the PR...

Unit test fails. There are more old files in the build tree...

OlofGullnas:
The failing unit tests use queues. I suspect my problem with the unit tests is my old issue of having an NFS mounted home directory. This causes sqlite databases to be on an NFS  mounted disk which is not good.

OlofGullnas:
I looked at the help for some build commands. I noticed a difference in --verbose. The build command for executable and server has this as an option without argument. casual-build-resource-proxy has this as an option with an argument (0/1/true/false...?)

There is also a difference in the option to keep intermediate file. For server and executable this is -k or --keep. For resource proxy it is -s or --keep-source. (-k is used for the resource key).

I also noticed that casual-build-resource-proxy complains with 

 `[casual:invalid_argument] resource-key:  not found: invalid_argument
`
after the help text when you do casual-build-resource-proxy --help. 

Backward compatibility is an obvious concern/issue when changing command options and may make it infeasible to use the new shared settings also for casual-build-resource-proxy.

OlofGullnas (approved):
Changes are reasonable.
I would have named things a little different. I would like to emphasize that some "settings" are command line options/settings that are common to several (potentially all) of the "build" commands. I.e. for servers, executables and resource prexies. Making the build of resource proxies more uniform with the other would break backward compatibility.

Some combination of the words "shared", "build" and {"settings" or "options"}. The word "common" should probably be avoided :-)

lazan:
> Some combination of the words "shared", "build" and {"settings" or "options"}. The word "common" should probably be avoided :-)

The _fqn_ is `tools::build::setting::Directive`. It's not _options_ per se, it's what argument::options sets/assign. I don't want to use the word _option(s)_.  Hmm, I can't come up with something that feels ok and better.

`tools::build::settings::Mandatory` it is.

PR #606, opened 2025-09-17
Patch/1.8/main

lazan:


documentation/changelog.md
```diff
@@ -34,6 +34,10 @@ This is the changelog for `casual` and all changes are listed in this document.
 ### Fixes
 - common: time_point is not logged with iso8601 utc offset in some places ([#552](https://github.com/casualcore/casual/issues/552))
```
ahhud, line 41: I think we need to have an entry for 1.8.1. I.e. what's the reason for the version?

ahhud (approved):

PR #609, opened 2025-09-26
buffer: casual_field_match does not "to_string" correctly

### LINKED ISSUE #608: casual_field_match does not seem to "to_string" correct internally for string values
It seems like it "prints" the pointer value  `FML_ARCHIVE_STRING[0] = 0x1d3a548`

lazan:
Before: internally string values was "to_string" as hex-pointer value. Artifact from refactoring buffer from `std::vector< char>` to `std::vector< std::byte>`

Now: internally we "to_string" as `const char*`

Resolves #608

### DIFF
--- a/middleware/buffer/source/field.cpp
+++ b/middleware/buffer/source/field.cpp
@@ -1887,7 +1887,7 @@ namespace casual
                               stream << local::decode< long>( data);
                               break;
                            case CASUAL_FIELD_CHAR:
-                              stream << data;
+                              stream << *reinterpret_cast< const char*>( data);
                               break;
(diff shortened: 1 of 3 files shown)


middleware/buffer/source/field.cpp
```diff
@@ -1896,7 +1896,7 @@ namespace casual
                               stream << local::decode< double>( data);
                               break;
                            case CASUAL_FIELD_STRING:
-                              stream << data;
+                              stream << reinterpret_cast< const char*>( data);
```
KristianIvarsson, line 1899: There's probabley a (double) defect in `CASUAL_FIELD_CHAR` as well

PR #613, opened 2025-09-30
archive: Introducing TOML archive in favour of INI archive

KristianIvarsson:
A new archive and buffer type for the TOML protocol and dispatches for both .toml and .ini files along with TOML and INI buffers


middleware/common/source/serialize/toml.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+#include <toml++/toml.hpp>
+
+namespace casual
+{
+   namespace common
+   {
+      namespace serialize
+      {
+         namespace toml
```
lazan: I think you should collapse these to `common::serialise::toml` 
And do the explicit template instantiation at the end as ` template struct  common::serialize::reader::create::Registration< toml::local::reader::Implementation>:` and so on.

middleware/common/source/serialize/toml.cpp
```diff
@@ -0,0 +1,388 @@
...
+                           {
+                              m_stack.push_back( &m_table);
+                              name = name ? name : "";
+                           }
+                           
+                           if( name)
+                           {
+                              auto node = m_stack.back()->as_table()->find( name);
+
+                              if(node != m_stack.back()->as_table()->end())
```
lazan: `if(<ws>...`

middleware/common/include/common/serialize/toml.h
```diff
@@ -36,9 +36,16 @@ namespace casual
...
+            {    
+               serialize::Reader reader( const std::string& source);
+               serialize::Reader reader( std::istream& source);
+               serialize::Reader reader( const platform::binary::type& source);
+            }
+
             serialize::Writer writer();
 
-         } // ini
+         } // toml
```
lazan, line 48: I think you should collapse the ns

middleware/common/include/common/serialize/service/protocol/implementation.h
```diff
@@ -57,48 +57,48 @@ namespace casual
...
-         static constexpr auto type() { return common::buffer::type::xml;}
+         static constexpr auto type() { return common::buffer::type::yaml;}
 
       private:
          common::serialize::Reader m_reader;
          common::serialize::Writer m_writer;
       };
 
-      struct Ini : public Base
+      struct Xml : public Base
```
lazan, line 96: Did you rearrange these based on alphabetical order?

lazan (approved):
Looking good. Hade a few minor suggestions.

PR #616, opened 2025-10-18
refactoring: improved and simplified terminal::format

lazan:
This patch: we use the _column-formatters_ parameter pack directly when calculating width and printing, with help of fold-expression. Now it's a function api that is easier to use.

### DIFF
--- a/middleware/domain/source/manager/admin/cli.cpp
+++ b/middleware/domain/source/manager/admin/cli.cpp
@@ -267,10 +267,14 @@ namespace casual
 
             namespace format
             {
-               template< typename P>
-               auto process()
+               auto processes( auto& processes)
                {
-                  auto format_configured_instances = []( const P& entity)
+                  auto format_alias = []( const auto& entity)
+                  {
+                     return entity.alias;
+                  };
+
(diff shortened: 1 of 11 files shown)


middleware/domain/source/manager/admin/cli.cpp
```diff
@@ -280,22 +284,22 @@ namespace casual
...
                   {
                      if( entity.restart) 
                         return "true";
                      return "false";
                   };
 
-                  auto format_state = []( const P& entity)
+                  auto format_state = []( const auto& entity)
                   {
                      if( ! entity.enabled)
```
KristianIvarsson, line 304: I'd prefer the opposite way around, i.e. `if( entity.enbabled) return "enabled"; return "disabled";`

lazan: Yeah, but then you have to have a large if-scope for _enabled_. I try to have early returns if it's possible. I think it's a lot easier to reason about.

KristianIvarsson (approved):
Nothing to comment about the new code, but a minor comment about related (old) code

PR #625, opened 2025-11-03
domain: if server exit during boot the task might get stuck

### LINKED ISSUE #624: When server exit during boot - 'group scale out' task might get stuck
When this happens, a lot of functionality will be _blocked_. Such as  `--shutdown`.

lazan:
Before: When a server instance exit during boot/startup and the server has `restart`. The state for the instance remains _scale-out_, but no restart will happen (due to boot/startup runlevel). This will render the task 'stuck', and a lot of things will not work (all stuff that also uses task to get their job done, since they're blocked by the stuck task.

Now: If we're in another runleveln than `running` we just set wanted 'state' of the instance to _linger_, which the task will consider to be done. The instance state becomes `error` in the cli, which is the correct state for the instance in this scenario.

Resolves #624

### DIFF
--- a/middleware/domain/unittest/source/simple_server.cpp
+++ b/middleware/domain/unittest/source/simple_server.cpp
@@ -13,6 +13,8 @@
 #include "common/environment.h"
 #include "common/signal.h"
 
+#include "casual/argument.h"
+
 #include "common/signal.h"
 
 namespace casual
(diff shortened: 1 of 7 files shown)


middleware/domain/unittest/source/simple_server.cpp
```diff
@@ -22,9 +24,24 @@ namespace casual
...
+            )
+
+         };
+
+         void run( Settings settings)
          {
-            
+            if( settings.terminate)
+            {
+               log::information( "terminating as per settings");
```
KristianIvarsson, line 41: Maybe "terminating according to settings" or "terminating according to configuration"?

KristianIvarsson: ... or maybe "killed by death"

ahhud (approved):

middleware/domain/unittest/source/simple_server.cpp
lazan: It was copilot that suggested it. Good enough, and it's only in unittest.

AlbsAgren (approved):

PR #632, opened 2025-11-06
http: make sure to accept the Accept header suitable for GET requests

### LINKED ISSUE #623: HTTP inbould should look for Accept header
If the `Accept` header is present in a HTTP call, reply should (one of) the content-type's listed in that header

In order to be backward compatible, casual should look for header `Content-Type` and make the reply according to that if `Accept` is missing

KristianIvarsson:
In a GET request there's typically no body and thus the Content-Type header is a bit ambigious but the Accept header is then the correct one in order to tell the server what you can handle

As for this commit, it doesn't expect multiple values, e.g. `Accept: application/xml,application/json`

Resolves #623

### DIFF
--- a/middleware/http/source/inbound/call.cpp
+++ b/middleware/http/source/inbound/call.cpp
@@ -35,8 +35,10 @@ namespace casual
                {
                   if( auto found = algorithm::find( header, "content-type"))
                       return protocol::convert::to::buffer( found->value());
+                  if( auto found = algorithm::find( header, "accept"))
+                      return protocol::convert::to::buffer( found->value());
 
-                  common::code::raise::error( code::bad_request, "content-type header is mandatory");                 
+                  common::code::raise::error( code::bad_request, "content-type or accept header is mandatory");
                }
             } // buffer
 


OlofGullnas:
Hm...
My gut reaction is that if an accept header is present it should be the first criteria (not fallback) for selecting format for the response.

Using the content-type of the request is more of a "heuristic" for determining the format expected by requestor. Along the lines "requestor sent me data of type X, so it is likely that the same format is acceptible for the response".

If the requestor specifies a format that can be handled, it should be used. 

Might be a little different if the accept header lists several formats/types. In that case it can be reasonable to use the format in content-type as input when selecting from the list. It is likely to be among those listed, but this is not certain.

Changing the order to use accept as priority may requre selecting from the list when/if multiple formats/types are listed in the header. I assume content-type always is a single value, multiple values would be very confusing :-)

There is some discussion of this in for example https://stackoverflow.com/questions/5661596/do-i-need-a-content-type-header-for-http-get-requests

Some people in that discussion describe using the content type in the request to decide how to respond as something "smart servers" do to "adopt to the requestor" capabilities. Mentions as an example servers that have json as default response format, but that switch to XML if the reuest uses XML format

KristianIvarsson:
> Hm... My gut reaction is that if an accept header is present it should be the first criteria (not fallback) for selecting format for the response.

That is absolutelly correct, that this should be the first to look at. The PR just solves the issue for potential GET-callers where it makes no sense to add a `Content-Type` header, but the `Accept` header is also valid in a POST-call

Maybe we should solve it for real then also handle multiple `Accept` formats?

PR #639, opened 2025-11-19
file: make sure to expand paths in order to catch some bad input

KristianIvarsson:
The changes implicitly adds some extra controls and as a consequence paths are represented as absolute paths that probably is better in cli

Some minor refactoring was also made in order to catch any possible std::bad_alloc when doing some lexically work

This is not really a defect, but the resulting behaviour if the mechanism is used in a bad manner results in stuff that might be hard to understand, so if 1.9 is not so far away, maybe it could be targeted to 1.9

The manager is also now a bit more stable in regards of certain exceptions thrown

### DIFF
--- a/middleware/file/source/manager/resource.cpp
+++ b/middleware/file/source/manager/resource.cpp
@@ -82,7 +82,7 @@ namespace casual::file::resource
                if( std::filesystem::exists( rushes))
                   std::filesystem::rename( rushes, request.path);
                else
-                  std::filesystem::remove_all( request.path);
+                  std::filesystem::remove( request.path);
             }
 
             void rollback( const Request& request)
@@ -90,7 +90,7 @@ namespace casual::file::resource
                const auto rushes = local::temporary( request.trid, request.path);
 
                if( std::filesystem::exists( rushes))
-                  std::filesystem::remove_all( rushes);
+                  std::filesystem::remove( rushes);
             }
 
             void mitigate( const Request& request)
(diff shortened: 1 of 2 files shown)


middleware/file/source/manager/resource.cpp
```diff
@@ -141,26 +144,28 @@ namespace casual::file::resource
...
                // push to work load
-               state.working.push_back( std::move( request));
+               state.working.push_back( std::move( query));
                // send the reply
                state.multiplex.send( request.process.ipc, std::move( reply));
             break; case code::busy:
                if( request.wait)
                   // push to wait list
-                  state.pending.push_back( std::move( request));
+                  state.pending.push_back( std::move( query));
```
lazan, line 168: What is this `break; case <label>` ? I don't like it. It makes the code harder to read. 
I assume this is "a pattern" to help get rid of unwanted _fallthroughs_? Since c++17 you get a warning if you don't use explicit `[[fallthrough]] ` attribute, which we should use instead of this.
I understand that this is not added in this patch, and I missed this in previous PR. Please don't use this "pattern".

lazan (approved):
I approve the changes in the patch, but I really wan't us to get rid of `break; case <label>`

lazan:
You need to rename your branch. You can't use `#` in branch names (in our build-pipeline). Look at the failing builds.

lazan (changes requested):
You need to rename your branch.

lazan:
Good commit log!

middleware/file/source/manager/resource.cpp
KristianIvarsson: It's a Herb Sutter pattern

Issue #638, opened 2025-11-19
casual file treats empty file name as a valid file


KristianIvarsson:
... but fails when commit:ing the work

KristianIvarsson:
Probably we could remove the call to `lexically_normal()` and use `std::filesystem::weakly_canonical()` that will make an empty path become currect working directory and verify that the file is not a directory (that is good for other purposes as well) and report an error if that is the case

The possible caveat is that `std::filesystem::weakly_canonical()` will be resolved from the managers _current working directory_ if the path is a relative path that might differ from the requesters _current working directory_ but that will result in problems anyway

Maybe we should demand an _absolute path_ always ?

PR #641, opened 2025-11-24
file: make sure to expand paths in order to catch some bad input

KristianIvarsson:
The changes implicitly adds some extra controls and as a consequence paths are represented as absolute paths that probably is better in cli

Some minor refactoring was also made in order to catch any possible std::bad_alloc when doing some lexically work

Solves issue #638

### DIFF
--- a/middleware/file/documentation/file.development.md
+++ b/middleware/file/documentation/file.development.md
@@ -29,7 +29,7 @@ Multiple access to the same (original) path within the same transaction if of co
 ...
 
    const auto path = casual::file::reserve( the_path);
-   std::ofstream{ path} << data;
+   std::ofstream{ path, std::ios::noreplace} << data;
 
 ...
    tx_commit();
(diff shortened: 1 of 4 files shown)


middleware/file/documentation/file.development.md
```diff
@@ -45,7 +45,7 @@ Multiple access to the same (original) path within the same transaction if of co
 ...
 
    const auto path = casual::file::reserve( the_path);
-   std::ofstream{ path} << data;
+   std::ofstream{ path, std::ios::in} << data;
```
lazan, line 48: How does this work? _output_ stream open as _input_, for reading?

KristianIvarsson: `std::ios::out` is implicit for std::ofstream and with `std::ios::in` the handle won't be opened if the file doesn't exist i.e. for update only

lazan (approved):
Good git log.

PR #642, opened 2025-11-27
Patch/1.8/some bugs

lazan:


ahhud (approved):

middleware/common/source/argument.cpp
```diff
@@ -315,7 +320,7 @@ namespace casual
                   if( depth)
                      *depth -= 1;
 
-                  if( option.names().active().empty() && ! option.names().deprecated().empty())
+                  if( option.names().is_deprecated())
```
KristianIvarsson, line 323: `std::format` can join a container of strings out-of-the-box (but maybe that doesn't work in current _supported_ g++)

middleware/service/include/service/manager/state.h
```diff
@@ -264,22 +264,33 @@ namespace casual
                         CASUAL_SERIALIZE( service);
                      )
                   };
-                  
+
+   
+                  struct Unset 
+                  {
+                     CASUAL_LOG_SERIALIZE(
+                        CASUAL_SERIALIZE( "Unset");
```
KristianIvarsson, line 272: I guess this is more for "visibility" in _human readable protocols_

middleware/service/source/manager/handle.cpp
```diff
@@ -55,6 +55,36 @@ namespace casual
       {
          namespace
          {
+            auto set_timer( const state::service::pending::deadline::Directive& directive, common::chronology::time_point now = platform::time::clock::type::now())
```
KristianIvarsson, line 58: Why do we have 2 time_points (kind of)?

middleware/service/source/manager/handle.cpp
```diff
@@ -55,6 +55,36 @@ namespace casual
...
+                  common::signal::timer::unset();
+               }
+            }
+
+            std::string service_name( const State& state, state::service::id::type service_id)
+            {
+               if( state.services.contains( service_id))
+                  return state.services[ service_id].information.logical_name();
+               else
+                  return "<unknown>";
```
KristianIvarsson, line 75: I'm not so fund of this string-literal, but since it is used in many many places, shouldn't we have it as a centralized constant somewhere ?

middleware/service/source/manager/handle.cpp
```diff
@@ -1050,7 +1081,7 @@ namespace casual
 
                   // we remove possible deadline first.
                   if( auto deadline = state.pending.deadline.remove( message.correlation))
-                     common::signal::timer::set( deadline.value());
+                     local::set_timer( deadline.value());
```
KristianIvarsson, line 1084: `*deadline` in order to be consistent with other code

PR #644, opened 2025-12-01
Feature/1.9/queue fan out

lazan:


middleware/configuration/include/configuration/model.h
```diff
@@ -786,21 +786,108 @@ namespace casual::configuration
...
+            struct Queue 
+            {
+               struct Target
+               {
+                  std::string queue;
+                  common::chronology::duration delay{};
+
+                  friend auto operator <=> ( const Target&, const Target&) = default;
+                  friend bool operator == ( const Target& lhs, const Target& rhs) = default;
+                  inline friend bool operator == ( const Target& lhs, std::string_view queue) { return lhs.queue == queue;}
```
KristianIvarsson, line 800: Why `inline` ?

Why not `const std::string&` as in other places and maybe a `const T&`with `std::is_convertible_v<T, std::string_view>` or such?

middleware/configuration/include/configuration/user.h
```diff
@@ -832,6 +831,68 @@ namespace casual
...
+
+               struct Group
+               {
+                  std::optional< std::string> alias;
+                  std::optional< std::vector< fanout::Queue>> queues;
+                  std::optional< std::string> note;
+
+                  CASUAL_CONST_CORRECT_SERIALIZE(
+                     CASUAL_SERIALIZE( alias);
+                     CASUAL_SERIALIZE( note);
```
KristianIvarsson: When using C++26 reflection, the serialization order will (probably) become different

middleware/configuration/source/model/transform.cpp
```diff
@@ -514,6 +514,48 @@ namespace casual
                         return result;
                      });
                   }
+
+                  if( source.fanout && source.fanout->groups)
+                  {
+                     result.fanout.groups = algorithm::transform( *source.fanout->groups, []( auto& group)
+                     {
+                        configuration::model::queue::fanout::Group result;
+                        result.alias = group.alias.value_or( "");
```
KristianIvarsson, line 523: Use `.value_or( {})` if you want default initialized objects instead of sprinkle the code with literals

middleware/configuration/source/model/transform.cpp
```diff
@@ -514,6 +514,48 @@ namespace casual
...
+                              result.instances = queue.instances.value();
+
+                           if( queue.targets)
+                           {
+                              result.targets = algorithm::transform( *queue.targets, []( auto& target)
+                              {
+                                 configuration::model::queue::fanout::Queue::Target result;
+                                 result.queue = target.queue;
+                                 if( target.delay)
+                                    result.delay = common::chronology::from::string( target.delay.value());
```
KristianIvarsson: Use `*target.delay` in order to be consistant

middleware/queue/include/queue/fanout/state.h
```diff
@@ -0,0 +1,293 @@
...
+#include <variant>
+
+namespace casual
+{
+   namespace queue::fanout
+   {
+      namespace state
+      {
+      
+         enum struct Runlevel : short
```
KristianIvarsson, line 29: This is not related to this PR, but it seems like these are the only runlevels for every manager and could it be a good idea to make something in common for these?

KristianIvarsson (approved):
Over all good clear gode, good naming and such, but some overall concept description is missing, so I guess some documentation is needed to describe the correlation between QM since this is not really a parallel standalone manager

middleware/configuration/include/configuration/user.h
lazan: Ahh, I'll fix.

middleware/configuration/include/configuration/model.h
lazan: `inline` to not break ODR. 
`std::string_view` I think I only needed this in unittests.

middleware/configuration/source/model/transform.cpp
lazan: I concur.

middleware/configuration/source/model/transform.cpp
lazan: I concur.

middleware/queue/include/queue/fanout/state.h
lazan: Yeah, perhaps. But there are benefits to have it 'local' also.

middleware/configuration/source/model/transform.cpp
lazan: Computer says no. Hmm, should it work? Is it a clang defect?

PR #651, opened 2025-12-14
http: Looking for header Accept along with Content-Type to deduce buffer

KristianIvarsson:
I'm not too happy with the implementation, but the added unit tests shows the intention

I also removed the casual-generic-application content-type-handling

I added a few buffer - and content types and removed the null buffer

If were gonna support null buffer we need to reinsert that and if we need the casual-generic-handling we need to reinsert that as well

I also fixed some bad makefile in http component

### DIFF
--- a/middleware/http/include/http/common.h
+++ b/middleware/http/include/http/common.h
@@ -65,21 +65,20 @@ namespace casual
          constexpr std::string_view x_octet = "application/casual-x-octet";
          constexpr std::string_view binary = "application/casual-binary";
          constexpr std::string_view json = "application/json";
+         constexpr std::string_view yaml = "application/yaml";
+         constexpr std::string_view toml = "application/toml";
          constexpr std::string_view xml = "application/xml";
          constexpr std::string_view field = "application/casual-field";
+         constexpr std::string_view order = "application/casual-order";
          constexpr std::string_view string = "application/casual-string";
-         constexpr std::string_view null = "application/casual-null";
+         constexpr std::string_view null = "application/casual-null"; 
 
          namespace convert
          {
-            namespace from
-            {
-               std::string buffer( std::string_view buffer);
-            } // from
-
             namespace to
             {
-               std::string buffer( std::string_view content);
+               std::string_view content( std::string_view buffer);
+               std::string_view buffer( std::string_view content);
             } // to
          } // convert
       } //protocol
--- a/middleware/http/source/common.cpp
+++ b/middleware/http/source/common.cpp
@@ -7,7 +7,6 @@
 
 #include "http/common.h"
 
-#include "common/algorithm.h"
 #include "common/algorithm/compare.h"
 #include "common/algorithm/container.h"
 #include "common/buffer/type.h"
@@ -15,6 +14,7 @@
 #include "common/log/line.h"
 
 #include "casual/buffer/field.h"
+#include "casual/buffer/order.h"
 #include "casual/buffer/string.h"
 
 namespace casual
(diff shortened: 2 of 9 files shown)


KristianIvarsson:
This needs g++14 (or a more C++23-compliant compiler/library)

KristianIvarsson:
I fixed the issues issued by CoPilot (3 real unit-test-blunders)

KristianIvarsson:
@ahhud The branch/pr needs g++14 (or some C++23 compliant compiler) and that is why the build why if fails, but that would be fine in the c-make-branch but since you were suppose to bring this branch into that one it becomes a catch 22 😉

ahhud:
Ok, I will fix it asap. 🙂

ahhud:
@KristianIvarsson @lazan I have rebased this PR with feature/1.9/main i.e. with imake build functionality and now this branch builds and all tests are ok. So all we need now is an ok from one of the selected reviewers. :-)

middleware/http/source/inbound/call.cpp
```diff
@@ -31,13 +31,37 @@ namespace casual
...
+                        std::views::transform( []( auto type){ return std::string_view{ common::string::trim( type)};}) |
+                        std::ranges::to< std::vector>();
+                     
+                     if( content)
+                     {
+                        if( algorithm::find( accepts, content->value()))
+                           return { content->value()};
+
+                        // Normal HTTP behaviour is to allow different content-type than what is accepted
+                        return {}; // ... but this system is not normal
```
lazan: Explain why we don't want/can support it? Instead of just saying "not normal".

middleware/http/source/inbound/call.cpp
```diff
@@ -31,13 +31,37 @@ namespace casual
...
+                  auto content = algorithm::find( headers, "content-type");
 
-                  common::code::raise::error( code::bad_request, "content-type header is mandatory");                 
+                  if( auto accept = algorithm::find( headers, "accept"))
+                  {
+                     auto accepts = 
+                        accept->value() |
+                        std::views::split(',') | 
+                        std::views::transform( []( auto type){ return std::string_view{ common::string::trim( type)};}) |
+                        std::ranges::to< std::vector>();
```
lazan: Why do you transform to a vector, you could keep it as a range, and only if you're returning _accepts_ you convert.

middleware/http/source/inbound/call.cpp
```diff
@@ -31,13 +31,37 @@ namespace casual
...
+                     
+                     if( content)
+                     {
+                        if( algorithm::find( accepts, content->value()))
+                           return { content->value()};
```
lazan, line 49: I find it hard to reson about memory ownership in this code. If you wanna keep it as is, can you please comment about why it's ok to return reference data?

middleware/http/include/http/common.h
```diff
@@ -65,21 +65,20 @@ namespace casual
...
-            namespace from
-            {
-               std::string buffer( std::string_view buffer);
-            } // from
-
             namespace to
             {
-               std::string buffer( std::string_view content);
+               std::string_view content( std::string_view buffer);
+               std::string_view buffer( std::string_view content);
```
lazan, line 81: Yeah, this is better.

middleware/http/source/inbound/call.cpp
```diff
@@ -79,7 +103,7 @@ namespace casual
             {
                auto lookup( const communication::ipc::inbound::Device& ipc, const std::string& service)
                {
-                  message::service::lookup::Request lookup{ local::handle( ipc)};
+                  message::service::lookup::Request lookup{ handle( ipc)};
```
lazan: Why remove of the `local` ns? I think it's still help to easy see that it's a local function in this TU, even if we're calling it from within the `local`.

middleware/http/source/inbound/call.cpp
```diff
@@ -384,5 +408,21 @@ namespace casual
...
+      {
+         auto type( const std::vector< header::Field>& headers) -> std::string_view
+         {
+            for( auto result : 
+               local::buffer::type( headers) | 
+               std::views::transform( []( auto content){ return protocol::convert::to::buffer( content);}) | 
+               std::views::filter( []( auto type){ return ! type.empty();}))
+            {
+               return result;
+            }
```
lazan: Feels a bit hackish with the for-loop. Could you not just ask the range if it's not empty and just return deref it?

middleware/http/source/common.cpp
```diff
@@ -92,85 +92,47 @@ namespace casual
...
-                  }
+                  constexpr std::size_t buffer_type = 0;
+                  constexpr std::size_t content_type = 1;
 
-                  namespace buffer
+                  template< std::size_t key_index, std::size_t value_index>
+                  auto find( const auto& key) -> std::string_view
                   {
-                     namespace type
+                     static const std::vector< std::tuple< std::string_view, std::string_view>> mapping
```
lazan: Use an `array` instead? 
Otherwise a quite nifty solution.

lazan (changes requested):
Overall good, but I have a few comments.

middleware/http/source/inbound/call.cpp
KristianIvarsson: Clever 😉

middleware/http/source/inbound/call.cpp
KristianIvarsson: Well, it is not reference data per say, but I'll give it a clearification

middleware/http/source/inbound/call.cpp
KristianIvarsson: Matter of opinion and I think it makes it more bloated and kind of obstructs the purpose of namespaces

middleware/http/source/inbound/call.cpp
KristianIvarsson: Clearify

middleware/http/source/common.cpp
KristianIvarsson: `std::array` of cource 👍

middleware/http/source/inbound/call.cpp
lazan: Its string_view, It cant be more reference than that?

OlofGullnas: I interprete the comment in line 34 to mean that the function returns a view (i.e. a "reference") to some storage owned/managed by the caller (presumably the "headers" argument passed by const & to the function).

I'm not overly fond of using string_view except in a local context. Simpler/safer to use regular string. Avoiding construction of string objects saves a few cycles, but is this significant?

middleware/http/source/inbound/call.cpp
OlofGullnas: Yes, it is a matter of opinion. I agree with lazan. I think including local increases clarity. "handle" is a common name in casual and explicitly indicating that this is a local "handle" increases clarity.
When I see "handle" in code I also ask myself if it is the verb handle, mening "to process", or the noun handle (something you can grab, handtag in swedish). It is usually obvious from context. And in casual it is frequently (but not always) the verb. In this case it seems to be the noun... In other code bases I work with it's almost always the noun.)
Casual also uses namespaces as part of the naming convention for functions so it frequently makes sense to include more name space qualification than strictly needed.

In fact in this case i might even have named local::handle something else. Along the lines of  "my_process_handle" or "my_process". This my be overly verbose, but have a tendency/habit to use long names.

middleware/http/source/inbound/call.cpp
OlofGullnas: What is the intended/desired behaviour?

I'm assuming the task of local::buffer::type is to determine the allowed formats for the "response" buffer. The return value is  a vector of string_view so it can be "multi-valued".
  
The rules I infer from this are:
o if content-type and accept headers are present
   o return value will be content-type if present in accept (one element in vector)
   o return value will be empty if content-type is not in accept
o if accept is present and content-type is missing, a vector of accepted formats is returned
o if content-type is present but accept missing, content-type is returned (one element in vector)
o if both content-type and accept is absent, an empty vector is returned

An empty return value vector probably means "unknown response format" (an error condition).

If content-type and accept headers are present the content-type must be listed in accept and will be used for the response format. 

if content-type is present it will be used for the response format (in the absence of accept header) 

It is only in the case that content-type header is missing and accept header is present that local::buffer::type can return a vector with more than one element. In this case a format must be selected from the list somewhere else. Possibly as the first element in the list that is supported.

I do not know if it is common that content-type is missing/not specified. 

Would it be reasonable to return the llist of accepted formats when the format given in content-type is not listed in accept? It is probably unusal that the caller uses a content-type that is not accepted for the response, but it is not impossible. There is no guarantee that any of the formats listed in accept is "supported" by casual.

My understanding is that historically casual always used the format of the request for the response (and ignored accept header).

middleware/http/source/inbound/call.cpp
OlofGullnas: I have not looked into this fully. Is the result of this the first format returned by local::buffer::type() that is supported?

I'm guessing that protcol::convert::to::buffer(content) checks for supported/recognized "formats". The loop can possibly be replaced by returning the first element (if any) in the result range that is the result of the "pipe".

I have not used ranges enough to know the idioms. I also do not know how/if "lazy evaluation" comes into play.

middleware/http/source/inbound/call.cpp
KristianIvarsson: 2 against 1 and I change to whatever your opinion is in a democratic way

middleware/http/source/inbound/call.cpp
KristianIvarsson: Now when looking at it, it looks really stupid and must be some leftover from some kind of iterative coding process

middleware/http/source/inbound/call.cpp
KristianIvarsson: For typical HTTP-GET requests, there is no body and thus no content and thus no `content-type`, but instead just a possible `accept`, i.e. this is a list of what I can handle as a response

In a HTTP-POST request, normally there's a body with content and thus a `content-type` (e.g. `image/jpeg`) but the caller can tell the callee that it can handle (accepts) the responce body in certain types (e.g. `application/xml; application/json)

Currently, `casual` cannot handle "mixed" types (e.g. json-input and xml-result) so that's why this extra checks exist but we're trying to be as relaxed as possible anyway

Yes, historically `casual` demanded the request to contain the `content-type` header even though there was no body (e.g. in a typical GET-request) pretty much to know how to send the reply and that is not really the HTTP-way to go

The value of `*/*` is valid in the `accept` header and means that "I accept everything" and maybe we should handle that and just pick a type and return it and in scenarious where `content-type` exists it is easy to pick, but if not ... what type shall we pick ? I choose to make `casual` consider that to be an error, but that could of course be changed 🤔

lazan (approved):

middleware/http/source/inbound/call.cpp
```diff
@@ -103,7 +129,7 @@ namespace casual
                         common::code::raise::error( common::code::xatmi::timeout, "timeout during lookup of service: ", lookup.service.name);
                      case Enum::idle:
                      {
-                        request.process = local::handle( ipc);
+                        request.process = handle( ipc);
```
OlofGullnas: Did you miss to add back the local:: here :-)

OlofGullnas (commented):
I think Kristian missed to restore one  local:: namespace qualifier for a local::handle( ipc) in http/inboundf/call.cpp.

Code is still cals the correct function, but for uniformity....

middleware/http/source/inbound/call.cpp
KristianIvarsson: I did

KristianIvarsson:
> I think Kristian missed to restore one local:: namespace qualifier for a local::handle( ipc) in http/inboundf/call.cpp.
> 
> Code is still cals the correct function, but for uniformity....

Fixed and push:ed now (if not already merged)

PR #653, opened 2025-12-16
file: Send correct information to TM during prepare

KristianIvarsson:
Without this fix casual-file-manager cannot be used with other RM's

### DIFF
--- a/middleware/file/source/manager/resource.cpp
+++ b/middleware/file/source/manager/resource.cpp
@@ -211,6 +211,17 @@ namespace casual::file::resource
 
             return result;
          }
+
+         namespace xa::reverse
+         {
+            auto type( const auto& request)
+            {
+               auto reply = common::message::reverse::type( request);
+               reply.resource = request.resource;
+               reply.trid = request.trid;
+               return reply;
+            }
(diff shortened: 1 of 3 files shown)


middleware/file/source/manager/resource.cpp
```diff
@@ -211,6 +211,20 @@ namespace casual::file::resource
...
+            {
+               auto type( const auto& request)
+               {
+                  auto reply = common::message::reverse::type( request);
+                  reply.resource = request.resource;
+                  reply.trid = request.trid;
+                  return reply;
+               }
+            } // reverse
+         } // xa
```
lazan: nitpick: `namespace xa::reverse` .
Otherwise, nice to have an abstraction for the 3 xa messages.

lazan (changes requested):
Nice! But we need a commit message that describe what the problem was, and how it was fixed. 
For me I find it easiest to write with the "convention":  
Before: bla bla
bla bla
Now: bla bla
bla bla

lazan (approved):
I still think the commit message doesn't tell me enough about what has been fixed... But, it could be worse...

Issue #654, opened 2025-12-17
casual-queue-manager state service blocks -> queue-group take long time -> queue-lookup will be pending.


lazan:
Scenario:
* a queue-group is enqueued with 1M messages in the same transaction
* the commit phase takes really long time (which might be expected)
* a call to queue-manager state service is done
** queue-manager sends out state-request to all queue-groups (and forward-groups)
** queue-manager block-wait until all state-replies has been received
* during this time queue-manager can't answer queue-lookups

This could lead to system degradation. 

We should make the state service _concurrent_ to mitigate this. We've already got the framework/abstractions in place for this. casual-domain-manager has a few services as _concurrent_.

PR #656, opened 2025-12-19
queue: manager state service should be concurrent

lazan:
Before: state service was sequential. If some group takes really long time (a lot of messages enqueued, then commit), and user has called state service, state service will block until the busy group is done. During this time no queue-lookup will be served -> This could halt other parts of the user casual environment.

Now: state service is concurrent, and will not block other messages.

Resolves: #654

### DIFF
--- a/middleware/manager/include/casual/manager/service/protocol.h
+++ b/middleware/manager/include/casual/manager/service/protocol.h
@@ -64,7 +64,7 @@ namespace casual
          {
             using argument_type = R;
 
-            void operator () ( const R& argument)
+            void operator () ( argument_type&& argument)
             {
                invoke::Result result;
 
--- a/middleware/queue/source/manager/admin/server.cpp
+++ b/middleware/queue/source/manager/admin/server.cpp
@@ -7,6 +7,7 @@
 
 #include "queue/manager/admin/server.h"
 #include "queue/manager/admin/services.h"
+#include "queue/manager/state.h"
 #include "queue/manager/transform.h"
 #include "queue/manager/handle.h"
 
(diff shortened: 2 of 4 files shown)


middleware/manager/include/casual/manager/service/protocol.h
```diff
@@ -64,7 +64,7 @@ namespace casual
          {
             using argument_type = R;
 
-            void operator () ( const R& argument)
+            void operator () ( R&& argument)
```
KristianIvarsson: Why the name R and why make it an universal reference?

middleware/queue/source/manager/admin/server.cpp
```diff
@@ -26,35 +27,136 @@ namespace casual
...
+                        // This could happen later, if the group is busy and then shuts down (seems unlikely,
+                        // but possible)
+                        auto error_callback = []( auto& ipc, auto& complete)  
+                        {
+                           auto reply = common::message::reverse::type_t< Message>{};
+                           reply.correlation = complete.correlation();
+                           ipc::device().push( std::move( reply));
+                        };
+
+                        return state.multiplex.send( group.process.ipc, Message{ process::handle()}, std::move( error_callback));
```
KristianIvarsson, line 49: No need to move a closure with no capture

middleware/queue/source/manager/admin/server.cpp
```diff
@@ -26,35 +27,136 @@ namespace casual
...
                   {
-                     return entity.state == decltype( entity.state())::running;
-                  });
-               };
+                     return [ &state]( auto& group)
+                     {
+                        // If we fail to send, we just push an empty reply to our inbound device.
+                        // This could happen later, if the group is busy and then shuts down (seems unlikely,
+                        // but possible)
+                        auto error_callback = []( auto& ipc, auto& complete)
```
KristianIvarsson: So the ipc parameter is not used in this scenario?

Why not const declare them if not using any non const members?

middleware/manager/include/casual/manager/service/protocol.h
lazan: `argument` needs to be non-const, some protocol mutates argument. It's not a _universal reference_, it's an rvalue ref.  User needs to move the argument to "return from the service".

middleware/queue/source/manager/admin/server.cpp
lazan: More a convention to always move.

ahhud (approved):

PR #671, opened 2026-01-29
queue - casual_queue_example_server unittest fixes

### LINKED ISSUE #662: Unittest casual_queue_example_server failures
I have seen several failures in my 1.9 cmake build in the test suite casual_queue_example_server.. I 
got failures of the following tests in this suite:

1075:casual_queue_example_server.enqueue_1_message__expect_1_message_on_queue
1076:casual_queue_example_server.enqueue_1_message_then_abort__expect_empty_queue
1078:casual_queue_example_server.dequeue_empty_queue__expect_empty_response
1079:casual_queue_example_server.dequeue_1_message
1080:casual_queue_example_server.enqueue_multiple_queues

Some other tests may also succeed for the wrong reason!

I investigated the first one (the rest seems to be the same root cause).
The output in LastTest.log is 
```
Running main() from /export/home/guo/c_1.9/conan/p/b/gtest6a9a7a1300340/b/src/googletest/src/gtest_main.cc
Note: Google Test filter = casual_queue_example_server.enqueue_1_message__expect_1_message_on_queue
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from casual_queue_example_server
[ RUN      ] casual_queue_example_server.enqueue_1_message__expect_1_message_on_queue/export/home/guo/c_1.9/casual/middleware/queue/unittest/source/test_example.cpp:122: Failure
Expected equality of these values:
  local::call_enqueue( buffer, "example.q1")
    Which is: -1
  0

/export/home/guo/c_1.9/casual/middleware/queue/unittest/source/test_example.cpp:129: Failure
Expected equality of these values:
  messages.size()
    Which is: 0
  1UL
    Which is: 1

<end of output>
Test time = 1202.77 sec
----------------------------------------------------------
Test Failed.
"casual_queue_example_server.enqueue_1_message__expect_1_message_on_queue" end time: Jan 19 17:03 CET
"casual_queue_example_server.enqueue_1_message__expect_1_message_on_queue" time elapsed: 00:20:02

```

I added a printout of tperrno and it is TPENOENT. I suspect that the service is not yet advertised when the tpcall is executed.
Adding a short delay (I did a sleep 100ms) before the tpcall makes the test succeed. Without the sleep it usually fails, but occasianally it succeeds.

The other failed tests seems to be very similiar. They all report a -1 return code from the tpcall.

I also suspect that a number of other tests in the same test suite may be passing for the wrong reason!

For example casual_queue_example_server, enqueue_non_existing_queue__expect_failure. This intentions seems to be to tell the server to enqueue a message to a non-existing queue, get an error and retun an error to the caller. The test code only tests the return code from tpcall being -1.  No check of tperrno. This may cause the test to succeed because of a TPENOENTRY instead of an error returned from the called service...

In my environment it is common for the test case to "hang" when it fails. In this situation the tesdt binary has exited, but all the casual processes remain. The domain manager has a parent pid of 1 (instead of the pid of the process executing the test binary). Killing the casual processes causes the test run to proceed (or finish when I execute a single test using ctest with a -R <testname> option). This is very likely a separate problem to the unit test failures.  Possibly related to shutdown of the domain before it is completely started. I hae not seen this "hang" when the test cases (at least in this test suite) succeed.

OlofGullnas:
Modified unit tests to wait for a called service to become available. It could happen that the test called service before it was advertised by the server. This caused occasionalk test failures.

Also added check of tperrno after (expected) failed tpcall. Some tests could succeed for the wrong reason. I.e a tpcall could fail with a tperrno of TPENOENT when it should have failed with a different tperrno. Tests only checked for tpcall failure, and did not verify the specific tperrno.

Fixes #662

### DIFF
--- a/middleware/queue/unittest/source/test_example.cpp
+++ b/middleware/queue/unittest/source/test_example.cpp
@@ -10,6 +10,7 @@
 #include "domain/unittest/manager.h"
 #include "queue/api/queue.h"
 #include "common/buffer/type.h"
+#include "service/unittest/utility.h"
 
 namespace casual
 {
@@ -74,6 +75,12 @@ domain:
                return tpcall( common::string::compose("casual/example/enqueue/", queuename).data(), buffer, len, &buffer, &len, 0);
             }
 
+            // wait for a service called via call_enqueue (above) to become available.
+            void wait_for_enqueue_service(const std::string& queuename)
+            {
+               using namespace std::string_literals;
+               casual::service::unittest::wait::until::advertised( "casual/example/enqueue/"s + queuename);
+            }
          } // <unnamed> 
 
       } // local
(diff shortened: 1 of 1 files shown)


lazan (approved):
Looking good.

OlofGullnas:
Pushed a commit with corrected spelling. Also added the missing tpfree() calls. No ither changes.

OlofGullnas:
Squashed the commits and pushed again. Reduce "commit noise" :-) This seems to be the Casual way of doing things, but a forgot to do a squash at first :-)

OlofGullnas:
rebased to feature/1.9/main (as of 2026-05-05, commit 7b1ab16) and pushed to github.

PR #673, opened 2026-02-15
Patch/1.8/some bugs

### LINKED ISSUE #672: Conversational - Conversational services do not work from Java with casual 1.8, worked with 1.6
Conversational services does not work from Java to casual 1.8. This might be a (unintended and/or undocumented?) difference in the gateway protocol between 1.6 and 1.8. This worked fine with Casual 1.6

Possibly related to duplex in the final message of the conversation having a value of "terminated". Possibly also other differences in handling of normal "end" of a conversation and how this is communicated to the the callee/initiator.

Problem detected by Patrik Westman/Bolagsverket. Bolagsverket still uses 1.6, They just tried Casual 1.8 in a test environment and got problems with conversational transactions from Java.

Problem probably also detected by Mathias (the same day!?) when a Java test failed.

Possibly related to the commit "server: cleaned up server::context handlers" (july 18th 2025). I have so far only found one place that sets duplex to "terminated" and that is in server/source/context.cpp (line . The code to set duplex to terminated was introduced in that commit.

In service::conversation::Context::receive( ...) it looks as if it's message.code.result being filled in that is used to detect that this is a conversation_send message that ends a conversation (i.e. after a tpreturn). The value in duplex is not relevant (is ignored) in this case.

Should protocol be extended with "terminated" as a value for duplex, or should the gatway continue to use an earler convention.?

"Terminated" is a reasonable value for duplex for a terminated conversation. The other values (send or receive) does not make sense.

Will try to find out more.

lazan:


OlofGullnas (approved):
Looks good. 

Is the plan to introduce "terminated" in 1.9?

Using a value like terminated in Duplex is cleaner than relying on the result code  to signal "end of conversation". 

I asked Patrik at Bolagsverket to use a build of Casual with this PR included to verify that it  resolved the problem they saw when upgrading 1.6 -> 1.8.

The previously failing operations using the Java interface now worked without problems.

lazan:
> Looks good.
> 
> Is the plan to introduce "terminated" in 1.9?
> 
> Using a value like terminated in Duplex is cleaner than relying on the result code to signal "end of conversation".
> 
> I asked Patrik at Bolagsverket to use a build of Casual with this PR included to verify that it resolved the problem they saw when upgrading 1.6 -> 1.8.
> 
> The previously failing operations using the Java interface now worked without problems.

Perfect. We release this as fast as possible.

PR #689, opened 2026-03-03
gateway: outbound does not respect service_call v1.4 -> v1.2

### LINKED ISSUE #686: casual 1.8.1 sends the wrong service call request message when protocol is 1.4
When calling casual java, using protocol v1.4, from casual 1.8.1 we get the following on the java side:
2026-02-23 12:13:27,520 WARNING [se.laz.casual.network.inbound.reverse.ReverseInboundExceptionHandler] (EE-ManagedExecutorService-default-Thread-1) casual reverse inbound exception caught: io.netty.handler.codec.DecoderException: se.laz.casual.network.CasualDecoderException: se.laz.casual.api.network.protocol.messages.exception.CasualProtocolException: Message type 3100 is not supported by protocol version VERSION_1_4 closing channel

 The outbound service call request handler is missing the v1.4 protocol branch:                                                                                                              
                                                                                                                                                                                              
  // Line 299-302 of handle.cpp (outbound)
  if( message::protocol::compatible< common::message::service::call::callee::Request>( connection->protocol()))
     tcp::send( state, connection->descriptor(), message);                    // v1.5 → type #3104
  else
     tcp::send( state, connection->descriptor(),
        message::protocol::transform::to< common::message::service::call::v1_2::callee::Request>( std::move( message)));  // v1.2 → type #3100

  There are only two branches: v1.5 and v1.2. When the remote peer (the Java connector) negotiates v1.4 protocol, the v1.5 check fails, and the code falls through to the v1.2 fallback. This
  sends a service_call_v2 (#3100) message instead of the expected service_call_v4 (#3102).

Casual should handle sending the right service call request message also for protocol version 1.3 and 1.4.

lazan:
Before:
If outbound concluded that the other domain was not compatible with v1.5 for `service_call` it sent `v1.2` of `service_call`, regardless. That is, ignored if the other domain expected `v1.4` protocol.

Now:
outbound sends `v1.4` if that is what the other domain wants.

Resolves #686

### DIFF
--- a/middleware/gateway/unittest/source/protocol/1.4/test_manager.cpp
+++ b/middleware/gateway/unittest/source/protocol/1.4/test_manager.cpp
@@ -11,9 +11,12 @@
 
 #include "common/message/transaction.h"
 #include "common/communication/tcp.h"
+#include "common/communication/instance.h"
 
 #include "domain/unittest/manager.h"
+#include "domain/message/discovery.h"
 
+#include "service/lookup.h"
 #include "service/unittest/utility.h"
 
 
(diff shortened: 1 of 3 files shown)


ahhud (approved):

OlofGullnas:
Hm...
Code looks very reasonable, BUT I built the branch and tried the unit tests...

In my system the new unittest  gateway_protocol_1_4_manager.outbound_call_with_trid_request__expect_1_4__request hangs, and it seems to be a repeatable hang. I will look around a little and see if I can see where and why it hangs-

The unittest hangs at line 191 in protocol/1.4/test_manager.cpp:

`auto request = communication::device::receive< casual::domain::message::discovery::Request>( device);
`
At least that is the line gdb points at when I do a "bt" and look upwards in the stack.

The unit test code hangs in the same place when I execute the test with CASUAL_LOG="(.*)" so it is possible to get a (good?) "log" of what happens. 

[foo.log](https://github.com/user-attachments/files/25915215/foo.log)

There is a lot going on :-) The log was copied when the test had been hanging for a while.

I do not have a log from a working test case to compare with!  I suspect a timing dependency. I use a single CPU virtual machine. It is not unusual that I have timing related uit test failures noyt seen by other users. 

Looking at the log it appears to me as if casual-domain-discovery does not know about the "mockup domain" when the lookup request is done. The registration probably happens after the lookup. 

I tried adding a short sleep between the code that does "conect as a _mockup domain to the gateway" and the call to casual::service::lookup() (sleep after line 183). Then the unittest executed without failure. I guess it takes time for casual-domain-discovery to get the "registration" representing the mockup domain.

I used a sleep of 1 second, but I suspect that is a lot more then actually needed. And using a sleep is very "hacky".

A totally foolprof/correct test should probably check if the "mockup domain" is "known" before attempting the lookup. A worst case scenario could be  someone debugging  "gateway code" and have it sitting at a breakpoint... :-)

middleware/gateway/unittest/source/protocol/1.4/test_manager.cpp
```diff
@@ -157,5 +160,121 @@ namespace casual
...
+      reverse:
+         outbound:
+            groups:
+               -  connections: 
+                     -  address: 127.0.0.1:7010
+            )");
+
+         // connect as a _mockup-domain_ to the gateway.
+         auto device = unittest::tcp::connect::in( "127.0.0.1:7010", message::protocol::Version::v1_4);
+         EXPECT_TRUE( device.connector().socket());
```
OlofGullnas, line 183: I added a "process::sleep(1s);" after this line to get the unit test to succeed in my system. It apparently takes some time for the "mockup domain" to become known to "casual-domain-discovery". In my single CPU VM the service lookup done below seems to happen before the mockup domain is known, causing the test to hang  in the call at line 191.

A sleep is a hacky solution....

OlofGullnas (changes requested):

OlofGullnas:
I tried replacing the proces::sleep(1s) with a call

`auto state = gateway::unittest::state();
`

This seems to influence timing enough to make the test succeed. Not completely unexpected as it introduces a roundrip to the gateway manager which forces a few context switches.

My original intention was to add a retry loop (e.g. using common::unittest::eventually::succeed()?) with a predicate checking for a connection being established. The predicate could be state.connections.size() > 0 or a search the connections vector for an entry for a specified remote.name. I do not know if "connections" is the best thing to check. Perhaps something else should be checked instead.

Could possibly be a utility to add to gateway::unittest.

lazan:
> My original intention was to add a retry loop (e.g. using common::unittest::eventually::succeed()?) with a predicate checking for a connection being established. The predicate could be state.connections.size() > 0 or a search the connections vector for an entry for a specified remote.name. I do not know if "connections" is the best thing to check. Perhaps something else should be checked instead.
> 
> Could possibly be a utility to add to gateway::unittest.

We got one. `unittest::fetch::until( unittest::fetch::predicate::outbound::connected());`. There are other similar predicate with count and name, if I remember correct. But almost always one wants to wait until all otbounds are connected.

But I'm not sure if it would totally fix it. I think the issue is that outbound connection needs to register with _discovery_, to trigger an actual discovery (so we can call over the outbound). 
No, I checked the code and we register the first thing we do when we are connected. It should work with the _fetch-until_ above.

lazan:
@OlofGullnas Could you test the updated unittest?

OlofGullnas:
will do

OlofGullnas:
Works OK in my system.

OlofGullnas (approved):

Issue #703, opened 2026-04-14
queue-group does not take uncomitted messages into account for _empty_.


lazan:
queue-group does not take uncomitted messages into account when we check if a queue is _empty_. This is not so good when we try to determine if we can remove queues triggered by updated configuration (also during startup).

PR #704, opened 2026-04-14
queue: _empty queue_ does not take uncommitted messages into account

lazan:
Before:
When queue-group checks if a current queue is empty, to know if it's ok to remove the queue, uncommitted messages was excluded. This resulted in queue-group trying to remove a queue that has FK to messages, and sqlite, correctly, said stop.

Now:
We include uncommitted messages to determine if a queue is _empty_ or not.

Resolves: #703

### DIFF
--- a/middleware/queue/unittest/source/test_queue.cpp
+++ b/middleware/queue/unittest/source/test_queue.cpp
@@ -12,6 +12,7 @@
 #include "queue/unittest/utility.h"
 #include "queue/common/queue.h"
 #include "queue/api/queue.h"
+#include "queue/group/queuebase.h"
 
 #include "common/array.h"
 #include "common/code/queue.h"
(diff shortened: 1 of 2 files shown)


middleware/queue/unittest/source/test_queue.cpp
```diff
@@ -2124,6 +2125,56 @@ PRAGMA page_size = 512;
...
+         auto scope = common::unittest::environment::scoped::variable( "CASUAL_UNITTEST_QUEUEBASE", path.string());
+
+         // add 'uncommitted' message to the 'zombie' queue
+         {
+            group::Queuebase queuebase( path);
+            queuebase.create( group::queuebase::Queue{ .name = "a"});
+            auto zombie = queuebase.create( group::queuebase::Queue{ .name = "zombie"});
+
+            auto message = queue::ipc::message::group::enqueue::Request{};
+            {
```
KristianIvarsson, line 2142: No need for bracing this

partickl (approved):

PR #706, opened 2026-04-21
buffer: added payload 0 size to pool gives `null` buffer handle

### LINKED ISSUE #705: http-inbound does not propagate out-of-band header for GET requests
casual_forward_service does not propagate http header fields as out-of-band buffer header-fields.

lazan:
Before:
When a `payload` with zero size was `adopt` or `insert` to buffer pool, we didn't get a unique address for the buffer.
* User can't really use the buffer.
* User can't access header fields for the buffer

Now:
We make sure to reserve one byte if the `payload` is empty before we add the `payload` to the buffer pool via `adopt` and `insert`

Resolves #705


partickl (approved):

lazan:
Merged manually

PR #710, opened 2026-04-28
http: inbound handles wildcards in accept header field

### LINKED ISSUE #708: Bug in casual-http-inbound regarding mandatory Accept-header?
In casual 1.9 one now need to add header Accept in http-calls to casual-http-inbound.
Is that the intended behaviour?

This curl returns 404:
```
$ curl --data '{}' -H "content-type:application/json"  http://localhost:8080/.casual/service/state
<html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.28.0</center>
</body>
</html>
```

And with this entry in casual.log
```
2026-04-22T20:03:54.250306+02:00|domainC|450efc5cd9524e93aea5f371e69b6ec2|38236|0x1ee2c58c0|nginx||||error|[http:Not Acceptable] caught error: invalid or missing 'content-type'/'accept' headers: Not Acceptable
```
This curl returns requested data:
```
$ curl --data '{}' -H "content-type:application/json" -H "Accept:application/json" http://localhost:8080/.casual/service/state
{"result":{"instances":{"sequential":[{"process":{"pid":38239,"ipc":"hN4I2GQmRv6chyotlh75fQ=="},"alias":"casual-gateway-manager","state":0},{"process":{"pid":38235,"ipc":"QNuKoADmRBy2cq/g3EwR/w=="},"alias":"casual-example-server","state":0},{"process":{"pid":38234,"ipc":"9cb1pOUHTaqkmOziro9+6g=="},"alias":"casual-queue-manager","state":0},{"process":{"pid":38233,"ipc":"rvMNwNLfQBycjVV04JAQhQ=="},"alias":"casual-transaction-manager","state":0},{"process":{"pid":38230,"ipc":"EbGI2C7UTQOylUigrPzVyQ=="}
...
}
```

I have a major problem building casual due to an os upgrade right now, so I can't rebuild and test on the absolute latest version of casual 1.9. So one possibility is that this is already fixed.

I have now managed to build on latest version of casual 1.9 and there is still the same behaviour.

KristianIvarsson:
Afore: An error was issued when all header media types contained wildcards

After: Wildcards are now handled for matching the best media - or buffer type

Resolves: #708

### DIFF
--- a/middleware/http/source/inbound/call.cpp
+++ b/middleware/http/source/inbound/call.cpp
@@ -13,13 +13,12 @@
 #include "common/communication/instance.h"
 #include "common/code/category.h"
 
-
 namespace casual
 {
-   using namespace common;
-
    namespace http::inbound::call
    { 
+      using namespace common;
+
       namespace local
       {
          namespace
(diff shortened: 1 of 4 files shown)


middleware/http/source/inbound/call.cpp
```diff
@@ -32,36 +31,57 @@ namespace casual
...
             {
                //! @returns string pieces (views) owned by the caller
-               auto type( const std::vector< header::Field>& headers) -> std::vector< std::string_view>
+               auto type( const std::vector< header::Field>& headers) -> std::string_view
                {
+                  //
+                  // HTTP behaviour is normally to allow different content-type than what is accepted 
+                  // ... but currently, the source and target buffer need to be of the same type
+
                   auto content = algorithm::find( headers, "content-type");
```
lazan: If we got content-type, why not just return early?

middleware/http/source/inbound/call.cpp
```diff
@@ -32,36 +31,57 @@ namespace casual
...
+                     else
+                     {
+                        auto rank = []( std::string_view value)
+                        {
+                           auto parts =
+                              value
+                              | std::views::split( '/')
+                              | std::views::transform( []( auto part){ return std::string_view{ part};});
+
+                           return std::ranges::count( parts, std::string_view{ "*"});
```
lazan, line 64: `"*"sv`

middleware/http/source/inbound/call.cpp
KristianIvarsson: Well, if there's also `Accept` but with different type(s), we want to check that somehow and report an error instead of returning something unexpected

lazan (approved):

PR #711, opened 2026-05-04
Gateway - Fix transform of conversation connect request from v1_2, and added unittest, issue 709. 

### LINKED ISSUE #709: Gateway does not handle duplex correctly when transforming connect_request between protocol versions
Seems as if the gateway does not properly handle "duplex" when transforming between protocol versions.  This affects conversational services. The observed effect was that an incoming conversation_connect_request_v2 with "duplex: receive" was transformed to a conversation_connect_request with "duplex: send". This causes incorrect handling of ownership of the conversation (which end that should  send/receive). As a conversationa session is strictly half duplex this is pretty fatal  when different protocol versions are used.

OlofGullnas:
Fixed the transform of conversation::connect::callee::request <--> conversattion::connect::v1_2::callee::request.

duplex field was not copied during the transformation. This resulted in failures when a connect request has the value "receive" in the duplex filed (i.e. wen the "connecting" party specifies TPSENDONLY in the tpconnect call, indicating that more data will be sent after the connect.

Also added a simple unittest. The test should probably be moved to a new file in the 1.9 stream. In this PR it was added to an existing file to simplify a future merge merge of patch/1.8/main to feature/1.9/main.

### DIFF
--- a/middleware/gateway/unittest/source/test_protocol.cpp
+++ b/middleware/gateway/unittest/source/test_protocol.cpp
@@ -8,6 +8,7 @@
 
 #include "gateway/message.h"
 #include "gateway/message/protocol.h"
+#include "gateway/message/protocol/transform.h"
 #include "gateway/documentation/protocol/example.h"
 
 #include "common/serialize/native/network.h"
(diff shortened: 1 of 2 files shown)


middleware/gateway/unittest/source/test_protocol.cpp
```diff
@@ -104,7 +105,69 @@ namespace casual
...
+      //
+      // To avoid introducing a new file in the midle of the change from casual_make to
+      // cmake the test has been added to this existing file.
+      //
+      // This particular test was triggered by a bug where one member ("duplex")
+      // of the conversation::connect::V1_2::callee::request was not copied 
+      // the current version of the message when transformed.    
+      TEST(gateway_protocol, transform_conversation_connect_to_from_v1_2)
+      {
+         common::message::conversation::connect::callee::Request con_req;
```
lazan: I think `reqeust` is a better name, the _conversation_ (or _connect_) is deduced from context. 
And, I really hate abbreviations 😄 

Also, to make it easier later/below in the unittest:

```cpp
using Request = common::message::conversation::connect::callee::Request;
using v1_2_Request = common::message::conversation::connect::v1_2::callee::Request;

Request request;
```

middleware/gateway/unittest/source/test_protocol.cpp
```diff
@@ -104,7 +105,69 @@ namespace casual
...
+         // Change value of duplex. The value "send" set by fill
+         // corresponds to the default initialization value of duplex.
+         // To be able to detect a missing "copy" of duplex we need a
+         // "non-default" value.
+         con_req.duplex = decltype(con_req.duplex)::receive;
+         // Transform to V1_2. The transform::to requires a rvalue reference,
+         // so the "transform source" may be "destroyed"
+         auto con_req_copy = con_req;
+         common::message::conversation::connect::v1_2::callee::Request con_req_v1_2 =
+            gateway::message::protocol::transform::to< common::message::conversation::connect::v1_2::callee::Request>( std::move(con_req));
```
lazan: You get rid of the temporary names if you just do a copy as you pass the argument. 
```cpp
auto v1_2_request = gateway::message::protocol::transform::to< v1_2_Request>( Request{ request});
```

lazan: You could argue that the _transform::to_ functions should take by value... But I did wan't to ensure that we didn't accidentally copy something large, like a buffer.

lazan (changes requested):
Nice work. Good test-case. 
I think you should change some names though 😄 
At least get rid of the unnecessary _name for a copy_.

lazan:
Also, squash the two commits one.

middleware/gateway/unittest/source/test_protocol.cpp
OlofGullnas: @lazan I guessed avoiding a copy of potentially large objects was the reason for requiring move semantics.

Pushed a new version with changed names. Is this better?

I hate squashing commits when I know somebody (Bolagsverket in this case) is using a commit that would be squashed. I know it is common practice in the casual community to squash and force push,...

I can certainly do a squash if you want me to.

lazan: > I guessed avoiding a copy of potentially large objects was the reason for requiring move semantics.

Exactly. But if you move into by-value will avoid copy.

lazan: > I can certainly do a squash if you want me to.

It's fine. I'll merge this.

OlofGullnas:
Changed variable names and got rid of a temporary variable holding a copy of the request before transformations.

PR #714, opened 2026-05-06
queue: treat queue names prefixed with . as hidden

AlbsAgren:
Did it for zombies too out of some sense of symmetry, don't know if reasonable?

### DIFF
--- a/middleware/common/include/common/service/type.h
+++ b/middleware/common/include/common/service/type.h
@@ -9,6 +9,7 @@
 
 #include "common/code/xatmi.h"
 #include "common/serialize/macro.h"
+#include "common/name.h"
 
 #include <string>
 #include <iosfwd>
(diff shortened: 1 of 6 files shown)


middleware/common/include/common/service/type.h
```diff
@@ -40,7 +41,7 @@ namespace casual
       namespace hidden
       {
          //! @returns true if the service `name` is _hidden_ (starts with `.`)
-         bool name( std::string_view service);         
+         using common::name::hidden::name;
```
lazan: Nice. But we should get rid of `service::hidden::name` all together, don't you think? I have no problem if you wan't to keep this though.

lazan (changes requested):
Good work. 
Check out co-pilots comments, I think you at least needs to adress the one with `UB`.

AlbsAgren:
Thanks, I'll have a look

lazan (approved):
Looking good. Do you squash it?

I would like to have more description in the comment also. Steal some of Copilots overview?

PR #737, opened 2026-07-22
tools: normalize arguments and behavior

lazan:
This patch:
* use a 'mandatory' build setting type for all build executables
  * resource-proxy
  * server
  * executable

This will force/help us to provide exactly the same options for common build directives and such

@OlofGullnas could you check this PR?


lazan:
Merged manually

PR #738, opened 2026-07-22
common: remove all thread related stuff

lazan:
With this patch, `casual` is totally thread free. Removed the last thread abstraction that was used in a unittest.

* removed all thread abstractions
* created a `unittest-tcp-server' that accept connections and just echo the received message -> used to test tcp abstractions
* removed the thread ns in signal, since the scope stuff has nothing to do with threads


middleware/common/unittest/source/communication/tcp/server/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+               } listen;
+
+               CASUAL_LOG_SERIALIZE(
+                  CASUAL_SERIALIZE( listen);
+               )
+            };
+
+            auto validate( Settings settings)
+            {
+               auto raise_if_empty = []( const auto& value, auto error)
```
KristianIvarsson, line 44: This lambda was a bit of an overkill

middleware/common/unittest/source/communication/tcp/server/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+                  {
+                     Trace trace{ "common::unittest::tcp::server::local::dispatch::echo"};
+
+                     if( auto connection = common::algorithm::find( state.connections, descriptor))
+                     {
+                        if( auto complete = common::communication::device::non::blocking::next( *connection))
+                        {
+                           log::debug( "received: ", complete);
+
+                           // we need to set the offset to 0.
```
KristianIvarsson, line 116: The "why" is more important

middleware/domain/source/manager/handle.cpp
```diff
@@ -689,7 +689,7 @@ namespace casual
                         if( message.state.deceased())
```
KristianIvarsson, line 689: The word `deceased` is for humans (or possible animals in certain contexts)

Pick another word

lazan: That's the nomenclature we have. We have `assassination`, `contract` and such.

